PR #124, opened 2022-09-11
Expose Array not CoreArray

### LINKED ISSUE #123: cubed `Array` vs `cubed.CoreArray`
Hi, I'm just a bit confused by what the difference is between the `Array` object defined in `cubed.array_api.array_object` and the `cubed.CoreArray` that is returned by `cubed.from_array` and `asarray`. `CoreArray` doesn't implement the numpy API, but `Array` isn't importable from `cubed`'s top-level namespace.

Is this deliberate or should things be changed so that the only type a user sees is `Array`?

TomNicholas:
Closes #123.

Most of the changes are either fixing a circular dependency (where `cubed.core.ops` now imports `cubed.array_api.Array`, which imports the ops again), or writing down some type hints.

I can add isinstance checks in tests if you want.

The type hints don't do anything yet, especially as there is no pre-commit or CI setup. I can add that and start gradually typing the codebase if you want too?

### DIFF
--- a/cubed/__init__.py
+++ b/cubed/__init__.py
@@ -12,25 +12,28 @@ try:
 except Exception:  # pragma: no cover
     __version__ = "unknown"
 
-from .core import (
+from .array_api import Array
+from .core.array import (
     Callback,
-    CoreArray,
     Spec,
     TaskEndEvent,
     compute,
+    measure_baseline_memory,
+    visualize,
+)
+from .core.ops import (
     from_array,
     from_zarr,
     map_blocks,
-    measure_baseline_memory,
     store,
     to_zarr,
-    visualize,
 )
+
 
 __all__ = [
     "__version__",
     "Callback",
-    "CoreArray",
+    "Array",
     "Spec",
     "TaskEndEvent",
     "compute",
--- a/cubed/array_api/__init__.py
+++ b/cubed/array_api/__init__.py
@@ -1,5 +1,9 @@
 __all__ = []
 
+from .array_object import Array
+
+__all__ += ["Array"]
+
 from .constants import e, inf, nan, newaxis, pi
 
 __all__ += ["e", "inf", "nan", "newaxis", "pi"]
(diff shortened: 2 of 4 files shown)


TomNicholas:
Huh actually this still isn't working - `from_array` is still returning `CoreArray`, because somehow it's being converted from an `Array` to a `CoreArray` inside `map_blocks` somewhere.

(Type hinting would probably spot this bug immediately, might be a good time to add some!)

Also the whole `CoreArray._new` return `Array` thing seems like an odd design choice - why not just construct using `Array.__init__`?

tomwhite:
Thanks for working on this @TomNicholas!

> Also the whole `CoreArray._new` return `Array` thing seems like an odd design choice - why not just construct using `Array.__init__`?

I was thinking that it might be useful in case we want to support another array implementation - perhaps a fuller NumPy array implementation - in the future. But I think we can cross that bridge when we come to it, so +1 to changing to `Array.__init__`. It looks like you've left some instances of `CoreArray._new`.

> The type hints don't do anything yet, especially as there is no pre-commit or CI setup. I can add that and start gradually typing the codebase if you want too?

I'm fine with that. (As long as we don't have to spend a lot of time fighting with mypy 😄)

tomwhite:
BTW I tried building the docs and it failed. I think `CoreArray` needs changing to `Array` in https://github.com/tomwhite/cubed/blob/main/docs/api.rst

tomwhite:
Thanks for the contribution @TomNicholas!

TomNicholas:
Wait hang on did you test that worked before you merged it? I was still
working on it!

tomwhite:
Oops! Sorry, I thought it was just waiting on the docs change (which now works).

What else is needed?

TomNicholas:
Sorry! No I often push things commit-by-commit. I'm used to xarray where everyone is quite conservative about merging things (we have a policy of one approving review reuiqred from another core dev before merging).

> What else is needed?

Well there are no tests that the correct type is returned yet. And I didn't yet work out why `map_blocks` was returning a `CoreArray` when given an `Array`, so that bug is probably still there. Which means this doesn't actually fix #123.

I can do a follow-up PR today though.

tomwhite:
That sound great - sorry again for the inconvenience!

PR #128, opened 2022-09-13
Expose Array not CoreArray (round 2)

TomNicholas:
Follow on from #124.

I'm trying to type the return values so that mypy can verify for me that the correct type is being returned.

I've successfully typed the array creation functions module, but when I got to the ops module I became confused as to how the class design was supposed to work.

You have 

```python
# core/array.py
class CoreArray
    def rechunk(self: ?, chunks) -> ?:
        from cubed.core.ops import rechunk

        return rechunk(self, chunks)


# array_api/array_object.py
def rechunk(x: ?, chunks, ...) -> ?:
    return ...
```

How do you want this to work? Your private subclass is defining public API here. Typing wise this requires that we make `rechunk` pass types through unchanged, so `CoreArray.rechunk` returns `CoreArray` and `Array.rechunk` returns `Array`. This seems not helpful for ensuring that `.rechunk` as seen by the user always returns an `Array`.

I've used `rechunk` as an example here but I think you have a similar typing problem for many of the functions in `ops.py`, and one of the `map_*` functions returning a `CoreArray` is ultimately what's causing `from_array` to return a `CoreArray` (i.e. the specific bug I reported in #123).

Is there some obvious way of rewriting this that would ensure we get the correct type back in the public API?

### DIFF
--- a/cubed/array_api/creation_functions.py
+++ b/cubed/array_api/creation_functions.py
@@ -16,7 +16,8 @@ from cubed.core import (
 from cubed.core.ops import map_direct
 from cubed.utils import to_chunksize
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Tuple, List, Union
+
 
 if TYPE_CHECKING:
     from .array_object import Array
@@ -24,7 +25,7 @@ if TYPE_CHECKING:
 
 def arange(
     start, /, stop=None, step=1, *, dtype=None, device=None, chunks="auto", spec=None
-) -> Array:
+) -> "Array":
     if stop is None:
         start, stop = 0, start
     num = int(max(np.ceil((stop - start) / step), 0))
@@ -56,7 +57,7 @@ def _arange(a, size, start, stop, step):
     return np.arange(blockstart, min(blockstop, stop), step)
 
 
-def asarray(obj, /, *, dtype=None, device=None, copy=None, chunks="auto", spec=None) -> Array:
+def asarray(obj, /, *, dtype=None, device=None, copy=None, chunks="auto", spec=None) -> "Array":
     a = obj
     from cubed.array_api.array_object import Array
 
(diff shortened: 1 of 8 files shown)


TomNicholas:
> Is there some obvious way of rewriting this that would ensure we get the correct type back in the public API?

Could maybe just have

```python
ChunkedArray = TypeVar("ChunkedArray", bound=CoreArray)  # either Array or CoreArray

def rechunk(x: ChunkedArray, chunks, ...) -> ChunkedArray:
    return type(x)(name, target, spec, plan)

**Comment by tomwhite** [comment]
Is the problem trying to satisfy mypy? It seems that the runtime types are all `Array`, as expected.

I wonder if it would help to move the `rechunk` method from `CoreArray` to `Array`?

**Comment by TomNicholas** [comment]
> Is the problem trying to satisfy mypy?

The problem was that I was unconvinced that this

> It seems that the runtime types are all Array, as expected.

was actually true. So I was thinking about how to design the classes in such a way that this was always true. Adding type hints was just to help me work out if this was true or not.

However I've just realised I was being thrown off by the fact that `Array.__repr__` misleadingly prints `"CoreArray"` :sweat_smile: (Now fixed.)

I think I'm now happier that I'm no longer going to randomly find my xarray object is wrapping a `CoreArray`.

> I wonder if it would help to move the rechunk method from CoreArray to Array?

Maybe - `rechunk` is weird because it's the only function in `ops.py` that might accept and return `CoreArray` objects in addition to `Array` objects. All the rest of the functions there are now type hinted to only accept and return `Array` objects, which is what we want in order to guarantee no surprises in public API. Now the bug is fixed this is less important though, and I think this can be merged as is.

File: cubed/array_api/creation_functions.py [hunk 1]
```diff
@@ -207,8 +215,8 @@ def _linspace(x, *arrays, size, start, step, endpoint, linspace_dtype, block_id=
     )
 
 
-def meshgrid(*arrays, indexing="xy"):
-    if len({a.dtype for a in arrays}) > 1:
+def meshgrid(*arrs, indexing="xy") -> Tuple["Array", ...]:
```
tomwhite, line 218: Can this be kept as `arrays`? I'm trying to follow the naming in the array API spec: https://data-apis.org/array-api/latest/API_specification/generated/signatures.creation_functions.meshgrid.html

cubed/core/array.py
```diff
@@ -23,6 +26,10 @@ def gensym(name="array"):
     return f"{name}-{sym_counter:03}"
 
 
+# type representing either a CoreArray or a public-facing Array
+ChunkedArray = TypeVar("ChunkedArray", bound="CoreArray")
```
tomwhite, line 30: Maybe name this `T_ChunkedArray` or similar? I been vaguely thinking it might be better to call `CoreArray` `ChunkedArray`, or it might be useful for something else in the future.

cubed/tests/test_array_api.py
```diff
@@ -93,6 +93,13 @@ def test_asarray_from_array(spec):
     assert_array_equal(b.compute(), np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]]))
 
 
+def test_from_array(spec):
```
tomwhite, line 96: This test belongs in `test_core.py` since the function it is testing (`from_array`) is not in the array API.

tomwhite (commented):
Thanks for fixing the `repr`s - this would definitely be confusing! (I had tried some isinstance tests to convince myself the right types were being returned before.)

This almost ready to go in - just a few naming things in the comments.

TomNicholas:
Great - fixed all those things.

tomwhite (approved):

PR #129, opened 2022-09-22
Fix `meshgrid` so it passes array API tests

tomwhite:
The array API tests failed: https://github.com/tomwhite/cubed/actions/runs/3102948470. This fixes that failure, and also formatting issues (by running `pre-commit run --all-files`).

Note that the array API spec for `meshgrid` says that the return type is `List` (https://data-apis.org/array-api/latest/API_specification/generated/signatures.creation_functions.meshgrid.html), so I fixed that too.

@TomNicholas, I will merge this unless you have any comments or suggestions.

### DIFF
--- a/cubed/__init__.py
+++ b/cubed/__init__.py
@@ -21,14 +21,7 @@ from .core.array import (
     measure_reserved_memory,
     visualize,
 )
-from .core.ops import (
-    from_array,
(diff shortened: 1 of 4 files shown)


TomNicholas:
Oh my bad! Merge away.

Issue #67, opened 2022-07-18
Supporting xarray.apply_ufunc


TomNicholas:
A very large proportion of xarray's code goes through [`xarray.core.computation.apply_ufunc`](https://github.com/pydata/xarray/blob/392a61484e80e6ccfd5774b68be51578077d4292/xarray/core/computation.py#L846), which is also [exposed publicly](https://docs.xarray.dev/en/stable/generated/xarray.apply_ufunc.html) for advanced users to wrap their own ufuncs.

Internally if a dask array is present then we call [`dask.array.apply_gufunc`](https://docs.dask.org/en/stable/generated/dask.array.gufunc.apply_gufunc.html) - perhaps cubed should expose a similar function in the same way that it exposes `map_blocks` etc.?

xref https://github.com/pydata/xarray/issues/6807

TomNicholas:
Another question I had is whether the fact that `xarray.apply_ufunc` allows users to execute arbitrary code inside their custom ufunc (we have one in a group project that calls out to some fortran!) does that break the assumptions behind the whole memory predictability thing?

tomwhite:
> perhaps cubed should expose a similar function in the same way that it exposes `map_blocks`

Yes, I think it would be possible to write a `apply_gufunc` in cubed. It's not a part of the array API so it would live somewhere else in the namespace (probably the top-level like `map_blocks`).

> Another question I had is about the fact that `xarray.apply_ufunc` allows users to execute arbitrary code inside their custom ufunc (we have one in a group project that calls out to some fortran!). Does that break the assumptions behind the whole memory predictability thing?

To some extent it does, but there are things we could do.

We could ask users to provide an (estimate) of the extra memory required, like `map_direct` [already does](https://github.com/tomwhite/cubed/blob/400dc9adcf21c8b468fce9f24e8d4b8cb9ef2f11/cubed/core/ops.py#L355-L357) (although that function is not publicly exposed at the moment). 

I've created some notebooks that help judge if the memory estimates are working in practice, see https://github.com/tomwhite/cubed/blob/main/examples/lithops-add-random-local.ipynb for example. So there's some tooling like this we could provide to make it easier to provide and check memory estimates.

TomNicholas:
Sounds good.

I would be interested in trying to implement `apply_gufunc` in cubed so I can learn more about the internals, but I probably won't have any time to work on it for the next 2 weeks, so if you have added it already by then then no worries!

tomwhite:
[...]

PR #149, opened 2023-02-17
Add `apply_gufunc`

tomwhite:
This is for #67

This is a cutdown version of the equivalent code in Dask. It passes a number of tests, but the main limitation is that it only supports a single output. I've got a number of ideas for removing this limitation (e.g. write one output at a time - in series), but I think this is a good start for supporting xarray.

@TomNicholas I'll leave this open for a few days in case you want to take a look!


TomNicholas:
Awesome @tomwhite ! 

My only comment is that when `xarray.apply_ufunc` is called with dask then `output_sizes` is a required kwarg (at least when creating new dimensions on the output IIRC). I would need to look at the implementation in xarray again, but it might be the case that `output_sizes` would be required for xarray to make use of `cubed.apply_gufunc` too.

tomwhite:
Thanks for taking a look @TomNicholas and pointing out that `output_sizes` is needed in xarray.

I'm going to merge this and create a follow-on issue for `output_sizes` (see #150), as it needs a bit more investigation.

tomwhite:
Support for `output_sizes` is in #151

PR #168, opened 2023-05-10
Optional dependency specification and pyproject.toml cleanup

TomNicholas:
[Looking at dask](https://github.com/dask/dask/blob/main/pyproject.toml) it seems everything can go in the `pyproject.toml`, and that's now recommended by PEP 518, so I think we should aim to move everything in there.

I tested these changes by running `pip install -e .[complete]` from the cloned cubed directory locally, which appears to work as hoped.


pyproject.toml
```diff
@@ -1,3 +1,56 @@
...
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.8",
+]
+requires-python = ">=3.8"
+dependencies = [
+    "aiostream",
+    "fsspec",
+    "mypy_extensions", # for rechunker
```
TomNicholas, line 26: Is this still required now that cubed doesn't depend on rechunker?

pyproject.toml
```diff
@@ -1,3 +1,56 @@
...
+    "numpy >= 1.22",
+    "pydot",
+    "tenacity",
+    "toolz",
+    "zarr",
+]
+
+[project.optional-dependencies]
+persist = ["dill"]
+diagnostics = ["tqdm", "graphviz"]
```
TomNicholas, line 37: `graphviz` vs `ipycytoscape` vs neither here?

pyproject.toml
```diff
@@ -1,3 +1,57 @@
...
+    "numpy >= 1.22",
+    "pydot",
+    "tenacity",
+    "toolz",
+    "zarr",
+]
+
+[project.optional-dependencies]
+persist = ["dill"]
+diagnostics = ["tqdm", "graphviz", "pandas"]
```
TomNicholas, line 37: `pandas` is here to facilitate examining the memory utilization

TomNicholas:
This seemed to work as intended, but now it fails because the test suite doesn't distinguish between tests which depend on different executors. Ideally we would have a suite of test runs, each of which tests just one executor, so you can easily see from the CI summary which are still green. That might require individual tests to be decorated with e.g. `requires_lithops` similar to how we do it in xarray. Alternatively I could just make `cubed[test]` include `cubed[complete]`, in which case all test environments will attempt to install all executors.

pyproject.toml
tomwhite: Yes, the vendored copy of rechunker uses it for `NamedArg`:

https://github.com/tomwhite/cubed/blob/74145fb826d21496ca86b4e64b0fcb1c55d8a931/cubed/vendor/rechunker/types.py#L15

Maybe we could remove that at some point, but it's needed as things currently stand.

pyproject.toml
```diff
@@ -1,3 +1,57 @@
...
+    "Programming Language :: Python :: 3.8",
+]
+requires-python = ">=3.8"
+dependencies = [
+    "aiostream",
+    "fsspec",
+    "mypy_extensions", # for rechunker
+    "networkx < 2.8.3",
+    "numpy >= 1.22",
+    "pydot",
```
tomwhite, line 29: I think we could move `pydot` to `diagnostics`, since it's only needed when you call `visualize`. It has a dependency on `graphviz`, which is already in `daignostics`.

tomwhite:
Thanks for this PR and for converting to pyproject.toml!

> Ideally we would have a suite of test runs, each of which tests just one executor, so you can easily see from the CI summary which are still green.

That would be ideal, however that could be done separately too if you added `apache-beam` and `lithops` (but not `gcsfs`) to run tests for the moment. Modal tests are currently separate since they have conflicting package requirements.

pyproject.toml
```diff
@@ -1,3 +1,57 @@
...
+    "tenacity",
+    "toolz",
+    "zarr",
+]
+
+[project.optional-dependencies]
+persist = ["dill"]
```
tomwhite, line 36: I'm not sure we need a separate group for this - it's a pretty obscure use case (serializing the state of a computation). So just add `dill` to `test`?

TomNicholas:
The test pass, so I think this is good to merge?

tomwhite (approved):
LGTM

Issue #166, opened 2023-05-08
Package with conda


TomNicholas:
The pangeo community primarily uses conda-forge, so we should put cubed up there.

tomwhite:
Good idea. I'm not very familiar with conda packaging, is this something you could get started - or give some pointers?

TomNicholas:
Cubed is [now available](https://anaconda.org/conda-forge/cubed) on conda - the [feedstock is here](https://github.com/conda-forge/cubed-feedstock).

My understanding is that now this will [automatically update](https://conda-forge.org/docs/maintainer/updating_pkgs.html) when a new release is uploaded to pypi. (We could automate releasing to pypi through github actions like xarray does, but that's another issue).

tomwhite:
Opened #179 to automate releases

PR #178, opened 2023-05-19
Docs on installing from conda + optional dependencies

TomNicholas:
Following #166, documents installing from conda and installing sets of optional dependencies.

Copied the idea of pointing to the actual dependencies specification file from the xarray getting started guide.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -2,11 +2,36 @@
 
 ## Installation
 
-Install Cubed with pip:
+### Conda
+
+You can install cubed with a minimal set of dependencies using conda:
+
+```shell
+conda install -c conda-forge cubed
+```
+
+### Pip
+
+You can also install cubed with pip:
 
 ```shell
 python -m pip install cubed
 ```
+
+Cubed has many optional dependencies, which can be installed in sets for different functionality (especially for running on different executors):
+
+    $ python -m pip install "cubed[diagnostics]"  # Install optional dependencies for cubed diagnostics
+    $ python -m pip install "cubed[beam]"         # Install optional dependencies for the beam executor
+    $ python -m pip install "cubed[lithops]"      # Install optional dependencies for the lithops executor
+    $ python -m pip install "cubed[modal]"        # Install optional dependencies for the modal executor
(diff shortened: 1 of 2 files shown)


docs/getting_started.md
```diff
@@ -2,12 +2,37 @@
...
 python -m pip install cubed
 ```
 
+Cubed has many optional dependencies, which can be installed in sets for different functionality (especially for running on different executors):
+
+    $ python -m pip install "cubed[diagnostics]"  # Install optional dependencies for cubed diagnostics
+    $ python -m pip install "cubed[beam]"         # Install optional dependencies for the beam executor
+    $ python -m pip install "cubed[lithops]"      # Install optional dependencies for the lithops executor
+    $ python -m pip install "cubed[modal]"        # Install optional dependencies for the modal executor
+    $ python -m pip install "cubed[complete]"     # Install all of the above
```
tomwhite, line 27: I just remembered that Beam, Lithops and Modal don't work together due to package conflicts, so we shouldn't have a `complete` option.

docs/getting_started.md
```diff
@@ -2,12 +2,37 @@
...
+
+To see the full list of which packages are installed with which options see `[project.optional_dependencies]` in `pyproject.toml`:
+
+.. literalinclude:: ../pyproject.toml
```
tomwhite, line 31: This didn't render properly when I tried it locally. Does it need something extra installed for Sphinx?

TomNicholas: Oh oops - I didn't clock that these docs files are .md not .rst :man_facepalming: 

How are you building these exactly? That's something else that should be added to the install instructions / CI.

tomwhite: CI runs https://github.com/tomwhite/cubed/blob/main/.github/workflows/docs.yml

To run locally you can create a python env with `./docs/requirements.txt` and then run `make clean html` in the `docs` directory.

TomNicholas: Thanks - turns out you can just wrap the rst command with `{eval-rst}`.

tomwhite (approved):

PR #180, opened 2023-05-20
Rename memory settings

### LINKED ISSUE #167: Rename memory settings
@TomNicholas said in https://github.com/tomwhite/cubed/issues/160#issuecomment-1533846337

> As an aside I find it a bit confusing that `max_mem` sometimes means "maximum memory the computation is allowed to use" (i.e. actual usage) and sometimes means "maximum memory the computation is projected to use" (i.e. a conservative estimate). For example the DAG visualization just shows me the number I entered as `max_mem`, which isn't particularly helpful here. It would be more helpful if it showed two numbers, i.e. `projected_memory_use / max_allowed_mem`, so I can see if I will exceed it or how much headroom I have.

See also #58

tomwhite:
Fixes #167 

* Deprecate `max_mem` in favour of `allowed_mem`, which now also includes `reserved mem`.
* Replace `required_mem` with `projected_mem`, which now also includes `reserved mem`.
* Update the documentation and add a new diagram showing the relationship between memory settings.


tomwhite:
I'd like to merge this soon unless you have any objections @TomNicholas

TomNicholas:
Looks good to me! Really like the new diagram.

PR #185, opened 2023-05-25
Reinstate Lithops AWS examples that were accidently removed in #174

tomwhite:


tomwhite:
Here's the timeline from running `lithops-matmul-random.py`:

1685454497_timeline

tomwhite:
> What code did you use to create these plots?

All you have to do is add `TimelineVisualizationCallback` - see the example in this PR. Code is in https://github.com/tomwhite/cubed/blob/main/cubed/extensions/timeline.py.

> Last week I started looking at this task timing information manually for the quadratic means example run with xarray, I'll try to post it later.

Great!

PR #188, opened 2023-05-26
Convert str to bytes

TomNicholas:
This is just a small quality-of-life improvement that allows passing memory arguments to `Spec` as strings with units, e.g. `allowed_mem='100MB'` rather than `allowed_mem=100_000_000`.

I went down a little rabbit hole learning about how there [isn't a particularly neat way](https://florimond.dev/en/posts/2018/10/reconciling-dataclasses-and-properties-in-python/) of doing input validation on a frozen `dataclass`, and ended up just removing the dataclass in favour of a normal class. 

The `Spec` object is now immutable. It could be mutable, like it technically was before, but I wasn't sure what the intended use was. If `Spec` is just intended as a container of kwargs then it should really be immutable (or even a `NamedTuple`, though that has similar problems with input validation).

### DIFF
--- a/cubed/core/array.py
+++ b/cubed/core/array.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass
 from operator import mul
-from typing import Optional, TypeVar
+from typing import Optional, TypeVar, Union
 from warnings import warn
 
 import numpy as np
@@ -8,7 +8,7 @@ from toolz import map, reduce
 
 from cubed.runtime.pipeline import already_computed
 from cubed.runtime.types import Executor
-from cubed.utils import chunk_memory
+from cubed.utils import chunk_memory, convert_to_bytes
 from cubed.vendor.dask.array.core import normalize_chunks
 
 from .plan import arrays_to_plan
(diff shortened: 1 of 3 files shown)


cubed/core/array.py
```diff
@@ -186,42 +186,107 @@ def __repr__(self):
...
 
-    def __post_init__(self):
-        if self.max_mem is not None:
-            warn(
-                "`max_mem` is deprecated, please use `allowed_mem` instead",
-                DeprecationWarning,
-                stacklevel=2,
-            )
-        if self.allowed_mem is None:
-            self.allowed_mem = (self.max_mem or 0) + self.reserved_mem
```
TomNicholas, line 224: Alternatively I could have just replaced this `__post_init__` with just

```python
    def __post_init__(self):
        if self.max_mem is not None:
            warn(
                "`max_mem` is deprecated, please use `allowed_mem` instead",
                DeprecationWarning,
                stacklevel=2,
            )
        object.__setattr__(self, 'reserved_mem', convert_to_bytes(self.reserved_mem))
        if self.allowed_mem is None:
            object.__setattr__(self, 'allowed_mem', (self.max_mem or 0) + self.reserved_mem)
        else:
            object.__setattr__(self, 'allowed_mem', convert_to_bytes(self.allowed_mem))
```

but that's (a) pretty hacky and (b) wouldn't pass a type checker :man_shrugging:

cubed/utils.py
```diff
@@ -157,3 +157,41 @@ def extract_stack_summaries(frame, limit=None):
...
+        elif size[-2:] in units and size[:-2].isdigit():
+            unit = size[-2:]
+            value = size[:-2]
+        else:
+            raise ValueError(f"Invalid value: {size}. Expected a string ending with an SI prefix.")
+
+        if unit in units and value.isdigit():
+            value = int(value)
+            # convert to bytes
+            return value * (1024 ** units[unit])
```
tomwhite, line 195: Powers of 10 are more common now, and are actually the [SI standard](https://en.wikipedia.org/wiki/Byte#Multiple-byte_units), so this should be 1000. This is what I did in https://github.com/tomwhite/cubed/blob/3742ac43ec318f5b322bda7b7e7ea003ac1a5513/cubed/utils.py#L51-L70

tomwhite (commented):
This looks very useful - thanks for adding it.

TomNicholas:
Great. This should be good to merge then.

tomwhite (approved):

PR #190, opened 2023-05-30
Improvements and fixes to Modal examples.

tomwhite:
Add matplotlib/seaborn dependencies to diagnostics to make timeline vizualisations work.


tomwhite:
Here's an example timeline when running `modal-matmul-random.py`
1685445298_timeline

Issue #164, opened 2023-05-06
Naming tasks by passing xarray metadata


TomNicholas:
I was looking deeper into how to make https://github.com/pydata/xarray/issues/7813 work.

So looks like the nodes are named when they are created by `Plan._new`. Q's:
- Would it make sense to name a task corresponding to array creation using the corresponding xarray `DataArray.name`? Other downstream tasks are composed from multiple arrays, so it's not clear how those should be named, but it seems reasonable to name initial nodes after the corresponding variable.
- Is it bad idea to have the naming scheme for a cubed DAG change depending on whether or not the DAG was created by xarray wrapping cubed or just by cubed on its own?
- Why does `from_array` create 3 nodes at once? Intuitively I would have thought that does creating a single array not just correspond to a single node? I can see that the stack goes down through `blockwise` but I don't understand why that means 3 nodes.
- If `from_array` had a `name` argument then I could pass `DataArray.name` to that from xarray. But the actual `Plan._new` call doesn't happen until way deeper - presumably we don't want to pass a `name` argument all the way down through `map_direct`, `map_blocks`, and `blockwise`?
- What's the distinction here between the name of the array and the name of the node? Is the name of the array really used for anything?

tomwhite:
> * Would it make sense to name a task corresponding to array creation using the corresponding xarray `DataArray.name`? Other downstream tasks are composed from multiple arrays, so it's not clear how those should be named, but it seems reasonable to name initial nodes after the corresponding variable.

I'm not sure how this works out with there being a 1:1 mapping from variables to cubed arrays, but the thing to do is to try it and see if it works.

> * Is it bad idea to have the naming scheme for a cubed DAG change depending on whether or not the DAG was created by xarray wrapping cubed or just by cubed on its own?

I don't think this is necessarily a problem. Also, rather than having xarray change the name of the cubed arrays it might just add some metadata that we could chose to display more prominently on the visualization of the DAG.

> * Why does `from_array` create 3 nodes at once? Intuitively I would have thought that does creating a single array not just correspond to a single node? I can see that the stack goes down through `blockwise` but I don't understand why that means 3 nodes.

[...]

PR #192, opened 2023-05-31
Show array variable name in visualization

tomwhite:
This change improves the DAG visualization by using the local variable names for labelling the nodes. This makes it a lot easier to see what is being computed.

I haven't thought about how this works in relation to Xarray, but we should be able to do something similar once the information is available from Xarray (#164).

This is an example from the tests:

```python
a = xp.asarray([[1, 2, 3], [4, 5, 6], [7, 8, 9]], chunks=(2, 2))
b = xp.asarray([[1, 1, 1], [1, 1, 1], [1, 1, 1]], chunks=(2, 2))
c = xp.add(a, b)
d = c * 2
e = c * 3

e.visualize(filename=tmp_path / "e")
```

Before:
before

After:
after

Note that you can still see the array names (`array-001` etc.) by hovering over the node to get the tooltip.

It would be nice to include constant expressions too (e.g. '2' and '3' above), but that could be added later.

Any thoughts @TomNicholas?

### DIFF
--- a/cubed/core/plan.py
+++ b/cubed/core/plan.py
@@ -248,10 +248,13 @@ class Plan:
                 first_cubed_summary = stack_summaries[first_cubed_i]
                 caller_summary = stack_summaries[first_cubed_i - 1]
 
-                display_name = array_display_names.get(n, "-")
+                if n in array_display_names:
+                    var_name = f" ({array_display_names[n]})"
+                else:
+                    var_name = ""
                 d[
                     "label"
-                ] = f"{display_name}\n{first_cubed_summary.name} {op_name_summary}"
+                ] = f"{n}{var_name}\n{first_cubed_summary.name} {op_name_summary}"
 
                 calls = " -> ".join(
                     [s.name for s in stack_summaries if not s.is_on_python_lib_path()]
--- a/cubed/core/plan.py
+++ b/cubed/core/plan.py
@@ -195,6 +195,25 @@ class Plan:
             "fontsize": "10",
         }
         dag.graph["node"] = {"fontname": "helvetica", "shape": "box", "fontsize": "10"}
+
+        # do an initial pass to extract array variable names from stack summaries
+        array_display_names = {}
+        for (n, d) in dag.nodes(data=True):
+            if "stack_summaries" in d:
+                stack_summaries = d["stack_summaries"]
+                first_cubed_i = min(
+                    i for i, s in enumerate(stack_summaries) if s.is_cubed()
+                )
+                caller_summary = stack_summaries[first_cubed_i - 1]
+                array_display_names.update(caller_summary.array_names_to_variable_names)
+        # add current stack info
+        frame = inspect.currentframe().f_back  # go back one in the stack
+        stack_summaries = extract_stack_summaries(frame, limit=10)
+        first_cubed_i = min(i for i, s in enumerate(stack_summaries) if s.is_cubed())
+        caller_summary = stack_summaries[first_cubed_i - 1]
+        array_display_names.update(caller_summary.array_names_to_variable_names)
+
+        # now set node attributes with visualization info
         for (n, d) in dag.nodes(data=True):
             if d["op_name"] == "blockwise":
                 d["style"] = "filled"
@@ -209,6 +228,7 @@ class Plan:
             target = d["target"]
             chunkmem = memory_repr(chunk_memory(target.dtype, target.chunks))
             tooltip = (
+                f"name: {n}\n"
                 f"shape: {target.shape}\n"
                 f"chunks: {target.chunks}\n"
                 f"dtype: {target.dtype}\n"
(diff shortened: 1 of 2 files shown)


cubed/core/plan.py
```diff
@@ -195,6 +195,25 @@ def visualize(
...
+                    i for i, s in enumerate(stack_summaries) if s.is_cubed()
+                )
+                caller_summary = stack_summaries[first_cubed_i - 1]
+                array_display_names.update(caller_summary.array_names_to_variable_names)
+        # add current stack info
+        frame = inspect.currentframe().f_back  # go back one in the stack
+        stack_summaries = extract_stack_summaries(frame, limit=10)
+        first_cubed_i = min(i for i, s in enumerate(stack_summaries) if s.is_cubed())
+        caller_summary = stack_summaries[first_cubed_i - 1]
+        array_display_names.update(caller_summary.array_names_to_variable_names)
```
TomNicholas, line 214: Can this all be refactored out into a function?

TomNicholas:
Oh that's really cool!

Using the actual python variable name was not an approach I knew was possible! I'm trying to think if that would lead to any weird situations...

Thoughts:
- Where could I pass the xarray objects' `.name` in here?
- Could this approach of naming after python variable names work with xarray objects wrapping cubed arrays? Would we have issues where it gets named "var" or something intermediate-level instead of after the top-level object the user actually created themselves?

cubed/core/plan.py
tomwhite: Yes, absolutely

TomNicholas:
Is there a possibility here of causing headaches whilst debugging? Like you want to know why your plan looks a certain way, so you define a new intermediate array variable so you can interrogate it, but now your plan looks different?

tomwhite:
> I'm trying to think if that would lead to any weird situations...

If you do some of the processing in a function, then call `visualize` outside the function (using arrays returned by the function, for example), then this still works. Also, you can even reuse variable names and it will be correct, even though the 1:1 mapping is broken (the underlying array names `array-001` etc ensure that there are no conflicts in the actual arrays).

> 
> Thoughts:
> 
> * Where could I pass the xarray objects' `.name` in here?

Not sure yet. We'll probably need to refactor to make that work.

> * Could this approach of naming after python variable names work with xarray objects wrapping cubed arrays? Would we have issues where it gets named "var" or something intermediate-level instead of after the top-level object the user actually created themselves?

I haven't tested that but it could be confusing (although the call stack is shown in the tooltip). We could exclude xarray from the modules that we extract array names from if it's a problem.

tomwhite:
> Is there a possibility here of causing headaches whilst debugging? Like you want to know why your plan looks a certain way, so you define a new intermediate array variable so you can interrogate it, but now your plan looks different?

The new plan would have the new intermediate array variable name though, wouldn't it?

TomNicholas:
> If you do some of the processing in a function, then call visualize outside the function (using arrays returned by the function, for example), then this still works. Also, you can even reuse variable names and it will be correct, even though the 1:1 mapping is broken (the underlying array names array-001 etc ensure that there are no conflicts in the actual arrays).

That sounds pretty neat.

> I haven't tested that but it could be confusing (although the call stack is shown in the tooltip). We could exclude xarray from the modules that we extract array names from if it's a problem.

Excluding xarray could work, but what if something else then wraps xarray? 

> The new plan would have the new intermediate array variable name though, wouldn't it?

I guess so, maybe that case is fine.

What happens with method chaining? Then you still technically have intermediate arrays but they don't get assigned their own python variable name do they? Would those arrays appear as '-' too?

---

I think this is a really cool idea, I'm just wondering if we might discover some reason later why libraries normally don't do this kind of stack introspection of user code... I guess if it's only used for visualization and not execution then it's pretty safe though.

tomwhite:
> I'm just wondering if we might discover some reason later why libraries normally don't do this kind of stack introspection of user code...

Another difference is that we are looking for a very specific type (`cubed.Array`), and not doing anything with general types.

> I guess if it's only used for visualization and not execution then it's pretty safe though.

Right, it's for display purposes.

BTW I found the technique here: https://stackoverflow.com/a/18425523

tomwhite:
> Where could I pass the xarray objects' `.name` in here?

The reason this is tricky is because the name is only assigned after the array is created. That's why in this PR the stack summary from the `visualize` call is needed (see `# add current stack info`). If it's left out then, in this example, `e` is not named correctly.

For Xarray, consider a call like the following from your [Means of quadratic quantities](https://gist.github.com/TomNicholas/8366c917349b647d87860a20a257a3fb) example:

```python
U = xr.DataArray(
    name='U',
    data=cubed.random.random((5000, 1, 987, 1920), chunks=(10, 1, -1, -1), spec=spec),
    dims=["time", "face", "j", "i"],
)
``` 

The Cubed array is created before the Xarray DataArray, so I'm not sure how Xarray would tell Cubed about the name.

So another idea I had is to expand the logic in this PR to cover Xarray DataArrays too - see if the underlying array is a Cubed array, and if it is then use the DataArray's name in the visualization. (This wouldn't introduce a dependency on Xarray - note we already have [a special case for Xarray arrays here](https://github.com/tomwhite/cubed/blob/6d37febd3928c92cd89c94a2cd45ea4b435c1a3c/cubed/array_api/creation_functions.py#L59)).

tomwhite:
035ccdb6076766bbf6cb8805689adb56a8f17ac0 shows the idea

Note the U and V are now correctly labelled (but there are lots of other arrays too!).

cubed-xarray

Also, the final node is not labelled UV, but it is if the example is updated to create a local variable (whose name is not actually used):

```
mean_UV = mean["UV"]
mean_UV.data.visualize("cubed-xarray")
```

But it might be possible to improve this too, by checking the argument to `visualize` to see if it's a DataArray.

tomwhite:
I've made a change so that the name is shown in parentheses when it's available. This should be lower risk and OK to merge.

e

PR #203, opened 2023-06-05
Add user guide and break out getting started into separate pages.

### LINKED ISSUE #201: Add a user guide
I'd like to add a user guide to the docs that explains the different executors, storage and memory settings etc.

tomwhite:
Fixes #201.

I would like to add a page about diagnostics (visualization, progress bar, timeline, etc), but will do that separately as this is a good improvement by itself I think.

cc @TomNicholas


docs/user-guide/memory.md
```diff
@@ -0,0 +1,48 @@
...
+
+Cubed provides a utility function, {py:func}`measure_reserved_mem <cubed.measure_reserved_mem>`, for measuring it for the runtime you are using. You can run this function first, then use the value it reports as the basis for setting `reserved_mem` for your computation. Cubed will use `reserved_mem` as a baseline when calculating `projected_mem`, in order to more accurately estimate the upper bound on memory usage, making it more likely that your computation doesn't exceed memory bounds.
+
+Some executors have the ability to measure memory as a task is running. They can report the `peak_measured_mem`, which is the highest actual memory usage for a given task during its execution. This information is useful for analysing how well the computation ran, and if there is room for improvement (in Cubed's algorithms for calculating `projected_mem`).
+
+## Chunk sizes
+
+As a rule of thumb `allowed_mem` should be at least ten times as large as your Zarr chunk sizes.
```
TomNicholas, line 39: Do your cubed chunk sizes have to be the same as your Zarr chunk sizes? And if not then what size chunk is likely to work well? Dask says 100MB as a general rule of thumb, but it might be different for cubed...

tomwhite: Cubed uses Zarr for all its array representations - so the chunk sizes are the same. If you started with a Zarr array with very large chunks then you might want to rechunk them to something smaller or better shaped before doing a reduction.

I think 100MB is probably a good rule of thumb - fast enough to retrieve/process/store, and also ~20 fit into typical allowed mem (2GB), which is good for reductions (as mentioned here).

TomNicholas: > I think 100MB is probably a good rule of thumb

Okay great, that's definitely worth adding (and also that 2GB is a reasonable `allowed_mem`).

> Cubed uses Zarr for all its array representations - so the chunk sizes are the same.

Right, I get the `cubed.Array` objects are backed by Zarr arrays of the exact same chunk sizes, but what about when you first call `cubed.from_array`? With dask it is possible to choose different chunks than those on disk, and sometimes there are good reasons to (e.g. combine multiple tiny chunks into 1 big one). Is that possible with cubed?

tomwhite: Yes, you can pass `chunks` to `from_array` and it will rechunk if necessary. See https://github.com/tomwhite/cubed/blob/7ccd99158bb17d23cd9ff8bc90841aff11c76685/cubed/core/ops.py#L33-L72. 

BTW I just noticed that this path doesn't have an Xarray fast path in the case that Xarray is backed by Zarr arrays. Is this a path that cubed-xarray uses?

Issue #208, opened 2023-06-08
Html repr for array


TomNicholas:
It would be nice to have a slick html repr for `cubed.Array` objects like exists for `dask.Array` objects. (If only so that when you have a cubed array with many chunks the repr isn't an enormous string)

This actually shouldn't be very hard, because I think practically everything could just be copied over from dask. It's basically just the [`_repr_html_` method here](https://github.com/dask/dask/blob/596d74dbc72db663efac606d42d5c93c5a917fb9/dask/array/core.py#L1627), which uses the [html template here](https://github.com/dask/dask/blob/596d74dbc72db663efac606d42d5c93c5a917fb9/dask/widgets/templates/array.html.j2). The template would need minor modifications. I think this should then just be [automatically detected by xarray here](https://github.com/pydata/xarray/blob/f4e0523be0ce8babaa8eff38365e5308b1fdb76b/xarray/core/formatting_html.py#L31).

PR #216, opened 2023-06-19
Add html array repr

TomNicholas:
Implements suggestion in #208.

99% of this PR is just vendoring the necessary dask functions.

Looks like this:
Screenshot from 2023-06-19 18-51-08

### DIFF
--- a/cubed/array_api/array_object.py
+++ b/cubed/array_api/array_object.py
@@ -1,4 +1,5 @@
 import operator
+import math
 
 import numpy as np
 
@@ -14,6 +15,11 @@ from cubed.array_api.dtypes import (
 from cubed.array_api.linear_algebra_functions import matmul
 from cubed.core.array import CoreArray
 from cubed.core.ops import elemwise
+from cubed.vendor.dask.utils import format_bytes, maybe_pluralize
+from cubed.vendor.dask.widgets import get_template
+
+
+ARRAY_SVG_SIZE = 120  # cubed doesn't have a config module like dask does so hard-code this for now
 
 
 class Array(CoreArray):
@@ -30,6 +36,49 @@ class Array(CoreArray):
 
     def __repr__(self):
         return f"cubed.Array<{self.name}, shape={self.shape}, dtype={self.dtype}, chunks={self.chunks}>"
+
+    def _repr_html_(self):
+        try:
+            grid = self.to_svg(size=ARRAY_SVG_SIZE)
+        except NotImplementedError:
+            grid = ""
+
+        if not math.isnan(self.nbytes):
+            nbytes = format_bytes(self.nbytes)
+            cbytes = format_bytes(math.prod(self.chunksize) * self.dtype.itemsize)
+        else:
+            nbytes = "unknown"
+            cbytes = "unknown"
+
+        return get_template("array.html.j2").render(
+            array=self,
+            grid=grid,
+            nbytes=nbytes,
+            cbytes=cbytes,
+            arrs_in_plan=f"{self.plan.dag.number_of_nodes()} arrays in Plan",
+            arrtype="np.ndarray",
+        )
+
+    def to_svg(self, size=500):
+        """Convert chunks from Cubed Array into an SVG Image
+
+        Parameters
+        ----------
+        chunks: tuple
+        size: int
+            Rough size of the image
+
+        Examples
+        --------
+        >>> x.to_svg(size=500)  # doctest: +SKIP
+
+        Returns
+        -------
+        text: An svg string depicting the array as a grid of chunks
+        """
+        from cubed.vendor.dask.array.svg import svg
+
+        return svg(self.chunks, size=size)
 
     # Attributes
 
--- a/cubed/array_api/array_object.py
+++ b/cubed/array_api/array_object.py
@@ -80,6 +80,12 @@ class Array(CoreArray):
 
         return svg(self.chunks, size=size)
 
+    def _repr_inline_(self, max_width):
+        """
+        Format to a single line with at most max_width characters. Used by xarray.
+        """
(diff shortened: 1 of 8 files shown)


cubed/array_api/array_object.py
```diff
@@ -31,6 +37,55 @@ def __array__(self, dtype=None) -> np.ndarray:
...
+        else:
+            nbytes = "unknown"
+            cbytes = "unknown"
+
+        return get_template("array.html.j2").render(
+            array=self,
+            grid=grid,
+            nbytes=nbytes,
+            cbytes=cbytes,
+            arrs_in_plan=f"{self.plan.dag.number_of_nodes()} arrays in Plan",
```
TomNicholas, line 58: Wasn't quite sure if this was the correct way to get the number of arrays in the plan.

cubed/array_api/array_object.py
```diff
@@ -31,6 +37,55 @@ def __array__(self, dtype=None) -> np.ndarray:
...
+            arrtype="np.ndarray",
```
TomNicholas, line 59: Because dask arrays can wrap chunks that are themselves other array types (e.g. sparse) this actually serves a purpose in dask, but right now for cubed this is hard-coded.

cubed/array_api/array_object.py
```diff
@@ -31,6 +37,55 @@ def __array__(self, dtype=None) -> np.ndarray:
...
+        """
+        from cubed.vendor.dask.array.svg import svg
+
+        return svg(self.chunks, size=size)
+
+    def _repr_inline_(self, max_width):
+        """
+        Format to a single line with at most max_width characters. Used by xarray.
+        """
+        return f"cubed.Array<chunksize={self.chunksize}>"
```
TomNicholas, line 87: Makes xarray's repr slightly better when collapsed, see https://docs.xarray.dev/en/stable/internals/duck-arrays-integration.html

TomNicholas: (Doesn't use the `max_width` kwarg but it's included for compatibility)

cubed/core/array.py
```diff
@@ -105,6 +107,16 @@ def size(self):
...
 
+    @property
+    def nbytes(self) -> T_IntOrNaN:
+        """Number of bytes in array"""
+        return self.size * self.dtype.itemsize
+
+    @property
+    def itemsize(self) -> int:
+        """Length of one array element in bytes"""
+        return self.dtype.itemsize
```
TomNicholas, line 118: Both of these are defined on dask arrays. I only strictly needed `nbytes`, but `itemsize` seemed useful too.

cubed/tests/test_html.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def test_repr_html():
+    pytest.importorskip("jinja2")
+    assert ones([])._repr_html_()
+    assert ones(10)[:0]._repr_html_()
+    assert ones(10)._repr_html_()
+    assert ones((10, 10))._repr_html_()
+    assert ones((10, 10, 10))._repr_html_()
+    assert ones((10, 10, 10, 10))._repr_html_()
```
TomNicholas, line 13: Also copied from dask. (Do we need to attribute it specifically?)

Surprisingly this seems to be the only test dask has for the array repr specifically.

cubed/array_api/array_object.py
tomwhite: This will give the number of arrays in the unoptimized plan. Maybe add a `num_arrays(optimize_graph=True)` method to `Plan` (to mirror the `num_tasks` method)?

cubed/core/array.py
```diff
@@ -105,6 +107,16 @@ def size(self):
         """Number of elements in the array."""
         return reduce(mul, self.shape, 1)
 
+    @property
+    def nbytes(self) -> T_IntOrNaN:
+        """Number of bytes in array"""
+        return self.size * self.dtype.itemsize
```
tomwhite, line 113: Will this ever be nan? Cubed doesn't have unknown size/shape arrays at the moment.

cubed/tests/test_html.py
tomwhite: > Also copied from dask. (Do we need to attribute it for license purposes?)

How about adding a comment referring to the license in vendor/dask.

tomwhite (commented):
Thanks for adding this! A few comments inline.

cubed/core/array.py
TomNicholas: I guess not.

cubed/array_api/array_object.py
```diff
@@ -31,6 +37,55 @@ def __array__(self, dtype=None) -> np.ndarray:
...
 
+    def _repr_html_(self):
+        try:
+            grid = self.to_svg(size=ARRAY_SVG_SIZE)
+        except NotImplementedError:
+            grid = ""
+
+        if not math.isnan(self.nbytes):
+            nbytes = format_bytes(self.nbytes)
+            cbytes = format_bytes(math.prod(self.chunksize) * self.dtype.itemsize)
```
tomwhite, line 48: (I thought I had made this comment in the review I did yesterday, but I can't find it now. Anyway, it's fine if we address it as a follow up if you want to merge this now.)

We should probably use Cubed's [`memory_repr`](https://github.com/tomwhite/cubed/blob/0f2fe5afff27789c78d80f48e6224dd30eed7779/cubed/utils.py#L51-L70) for consistency with the DAG visualization.

tomwhite (approved):

cubed/array_api/array_object.py
tomwhite: Thanks for fixing!

tomwhite:
We can merge this now if you want. (I can fix any problems that #218 shows up afterwards.)

PR #236, opened 2023-06-28
Add type hints for primitive functions

### LINKED ISSUE #230: Add type hints to blockwise and rechunk primitives
These would benefit from type hinting as an aid to understanding, and also to find bugs. This is especially true of the blockwise code, although it will be hard to specify good hints as there are a lot of nested structures and higher-order functions (this might be useful: https://peps.python.org/pep-0612/).

tomwhite:
Fixes #230

I'm interested if you have any feedback on this @TomNicholas

### DIFF
--- a/cubed/primitive/blockwise.py
+++ b/cubed/primitive/blockwise.py
@@ -1,13 +1,15 @@
 import itertools
+import math
 from dataclasses import dataclass
 from functools import partial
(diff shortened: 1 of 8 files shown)


tomwhite:
I'm going to merge this as it's a big improvement, even though there's more to do.

Issue #239, opened 2023-06-28
Flow control for worker task submission


tomwhite:
For very large computations when the number of tasks for an array is much greater than the number of workers, it may be desirable to have more control over task submission in the client, so a large number of submissions doesn't overwhelm the client  (memory) or the backend service.

Lithops' [`map`](https://lithops-cloud.github.io/docs/source/api_futures.html#lithops.executors.FunctionExecutor.map) function, for example, will materialize all the inputs before submission. So in this case it might be appropriate to batch the submissions. Note that Lithops' [`wait`](https://lithops-cloud.github.io/docs/source/api_futures.html#lithops.executors.FunctionExecutor.wait) function can wait on futures from different map submissions, so it should be possible to implement flow control this way.

For Modal, which has an async interface, it may be possible to just stream the inputs as they are produced on the client - it would need some experimentation.

PR #240, opened 2023-06-28
Test that plan generation scales to very large workloads

tomwhite:
Changes the `output_blocks` function to an iterable so that the entire list of chunk coordinates isn't materialized during planning. Without this change the test wouldn't complete in any reasonable time; with the change it takes a couple of seconds.

Of course, these huge jobs won't run since currently the entire iterable is materialized before being submitted to the backend. See #239.


tomwhite:
BTW @TomNicholas and @keewis this should fix the problem with saw with generating a plan for large workloads from the meeting on Monday.

Issue #224, opened 2023-06-22
Running on a cluster


TomNicholas:
It should be possible to try running Cubed on a cluster instead of on a serverless framework. Then the cluster just executes the tasks defined by the plan.

Writing a [dask executor for cubed](https://github.com/tomwhite/cubed/issues/161) would be one way to achieve this, but another might be to use [Lithops' Standalone Mode](https://github.com/lithops-cloud/lithops/blob/master/docs/source/execution_modes.rst#standalone-mode).

tomwhite:
Closed by #241

PR #241, opened 2023-06-28
Dask delayed executor

### LINKED ISSUE #161: Dask executor?
Would it make sense to add a dask executor? 

[Rechunker has one](https://github.com/pangeo-data/rechunker/blob/master/rechunker/executors/dask.py)

cc @rabernat

TomNicholas:
Would close #161. Also an example of #224.

The idea of this executor is that instead of guaranteeing that each instance doesn't exceed `allowed_mem` because it only sees one function to run (i.e. serverless), we are relying on dask.distributed's workers to each fully execute one delayed func before starting another (i.e. to not fall foul of [root task overproduction](https://www.coiled.io/blog/reducing-dask-memory-usage)). The embarassingly parallel nature of the list of `dask.Delayed` objects is what should hopefully keep the scheduler behaving predictably. Then the peak memory usage of the dask cluster *should* be just `n_workers` * `projected_mem`.

I haven't properly tried this yet, I just wanted to push this as a WIP, because it seems to work for a simple test case

Screenshot from 2023-06-28 12-47-32

I literally just copied how the sequential python executor works but wrapped `exec_stage_func` in `dask.delayed`. Currently this means that a new list of `dask.Delayed` objects is created then computed for each step in the cubed pipeline. I think the more advanced implementation would look more like [rechunker's dask executor](https://github.com/pangeo-data/rechunker/blob/master/rechunker/executors/dask.py), which builds each layer then creates a HLG before anything is computed.

### DIFF
--- a/cubed/runtime/executors/dask.py
+++ b/cubed/runtime/executors/dask.py
@@ -1,0 +1,37 @@
+import dask
+
+from cubed.core.array import TaskEndEvent
+from cubed.core.plan import visit_nodes
+from cubed.runtime.types import DagExecutor
+
+
+def exec_stage_func(func, *args, **kwargs):
+    # TODO would be good to give the dask tasks useful names
+    return dask.delayed(func(*args, **kwargs))  # should we add pure=True?
+
+
+class DaskDelayedExecutor(DagExecutor):
+    """Executes each stage using dask.Delayed functions."""
+
+    def execute_dag(self, dag, callbacks=None, array_names=None, resume=None, **compute_kwargs):
+        # Note this currently only builds the task graph for each stage once it gets to that stage in computation
+
+        for name, node in visit_nodes(dag, resume=resume):
+            pipeline = node["pipeline"]
+            for stage in pipeline.stages:
+                if stage.mappable is not None:
+                    stage_delayed_funcs = []
+                    for m in stage.mappable:
+                        delayed_func = exec_stage_func(stage.function, m, config=pipeline.config)
+                        stage_delayed_funcs.append(delayed_func)
+                        if callbacks is not None:
+                            event = TaskEndEvent(array_name=name)
+                            [callback.on_task_end(event) for callback in callbacks]
+                else:
+                    delayed_func = exec_stage_func(stage.function, config=pipeline.config)
+                    stage_delayed_funcs = [delayed_func]
+                    if callbacks is not None:
+                        event = TaskEndEvent(array_name=name)
+                        [callback.on_task_end(event) for callback in callbacks]
+
+                dask.persist(stage_delayed_funcs, **compute_kwargs)
--- a/cubed/runtime/executors/dask.py
+++ b/cubed/runtime/executors/dask.py
@@ -14,10 +14,10 @@ class DaskDelayedExecutor(DagExecutor):
     """Executes each stage using dask.Delayed functions."""
 
     def execute_dag(
-        self, dag, callbacks=None, array_names=None, resume=None, **compute_kwargs
+        self, dag, callbacks=None, array_names=None, resume=None, client=None, **compute_kwargs
     ):
-        # Note this currently only builds the task graph for each stage once it gets to that stage in computation
 
+        # Note this currently only builds the task graph for each stage once it gets to that stage in computation
         for name, node in visit_nodes(dag, resume=resume):
             pipeline = node["pipeline"]
             for stage in pipeline.stages:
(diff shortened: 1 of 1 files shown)


TomNicholas:
This did not behave the way I expected it to when I ran with a `LocalCluster` :confused: 

Screenshot from 2023-06-28 14-14-08

cubed/runtime/executors/dask.py
```diff
@@ -0,0 +1,48 @@
+import dask
+
+from cubed.core.array import TaskEndEvent
+from cubed.core.plan import visit_nodes
+from cubed.runtime.types import DagExecutor
+
+
+def exec_stage_func(func, *args, **kwargs):
+    # TODO would be good to give the dask tasks useful names
+    return dask.delayed(func(*args, **kwargs))  # should we add pure=True?
```
tomwhite, line 10: Looks like you are invoking the function when `exec_stage_func` is called?
```suggestion
    return dask.delayed(func)(*args, **kwargs)  # should we add pure=True?
```

TomNicholas: OOPS

TomNicholas: I'm still seeing spilling

Screenshot from 2023-06-28 16-14-12

cubed/runtime/executors/dask.py
```diff
@@ -0,0 +1,48 @@
...
+                else:
+                    delayed_func = exec_stage_func(
+                        stage.function, config=pipeline.config
+                    )
+                    stage_delayed_funcs = [delayed_func]
+                    if callbacks is not None:
+                        event = TaskEndEvent(array_name=name)
+                        [callback.on_task_end(event) for callback in callbacks]
+
+                client.persist(stage_delayed_funcs, **compute_kwargs)
```
tomwhite, line 48: Any reason you are calling `persist` rather than `compute`? Also should you unpack the args (`*stage_delayed_funcs`)?

(Not sure if either would make a difference though.)

TomNicholas: Persist because I have no need to collect all the results onto the scheduler process. https://stackoverflow.com/questions/41806850/dask-difference-between-client-persist-and-client-compute

Arg unpacking is a good point, let me try that.

TomNicholas: Persist because I have no need to collect all the results on the scheduler (see https://stackoverflow.com/questions/41806850/dask-difference-between-client-persist-and-client-compute), so I thought that was safer.

Unpacking the args is a good point, but I just tried it and I still see spilling.

tomwhite: > Persist because I have no need to collect all the results onto the scheduler process.

I think the function returns `None`, so no data will be transferred back anyway. May be worth trying `compute` if you've run out of other things to try...

TomNicholas:
Okay I was doing something monumentally stupid :sweat_smile: - I had rushed to create a notebook to test this idea but forgot to change the chunked array type from dask to cubed! So I was accidentally calling `.compute()` on a dask array. (For some reason dask didn't mind that I was passing an argument called `executor` to `dask.array.Array.compute()`...)

TomNicholas:
Here's what actually happens, I get an error from within Zarr

https://gist.github.com/TomNicholas/b1ce906e85fa7245bf85484479f609a0

cubed/runtime/executors/dask.py
```diff
@@ -0,0 +1,48 @@
...
+                else:
+                    delayed_func = exec_stage_func(
+                        stage.function, config=pipeline.config
+                    )
+                    stage_delayed_funcs = [delayed_func]
+                    if callbacks is not None:
+                        event = TaskEndEvent(array_name=name)
+                        [callback.on_task_end(event) for callback in callbacks]
+
+                client.persist(*stage_delayed_funcs, **compute_kwargs)
```
tomwhite, line 48: I ran your notebook locally with the following change and it worked. The memory was stable at around 2GB or less and there was no spilling to disk.
```suggestion
                dask.compute(*stage_delayed_funcs, **compute_kwargs)
```

I used the latest Dask Distributed and Xarray releases.

cubed/runtime/executors/dask.py
```diff
@@ -0,0 +1,48 @@
...
+            pipeline = node["pipeline"]
+            for stage in pipeline.stages:
+                if stage.mappable is not None:
+                    stage_delayed_funcs = []
+                    for m in stage.mappable:
+                        delayed_func = exec_stage_func(
+                            stage.function, m, config=pipeline.config
+                        )
+                        stage_delayed_funcs.append(delayed_func)
+                        if callbacks is not None:
```
tomwhite, line 36: Remove the callback handling here and below since it needs to be hooked in to Dask's API for delivering progress information (can be done later).

tomwhite (approved):
Thanks for adding this new executor!

A unit test would be good. We probably don't want to add `dask` to the regular tests in case we inadvertently start depending on it from the main code. So fine to do in a follow-up issue.

PR #249, opened 2023-07-05
Re-enable rechunker's `consolidate_reads` 

### LINKED ISSUE #235: Performance regression - `max_workers` not being used
I'm pretty sure there was a performance regression in the last few weeks, which appears for the quadratic means workload on the timeline visualization:

Before:
image

After:
image

Other things were changed too so perhaps ignore other aspects of this plot that are different, just notice that for some steps there is a group of tasks (exactly 334 out of 5334) that complete as a separate bunch later on. Looking at this again makes me wonder if this is actually instead some issue with the visualization...

@tomwhite you said you had an idea which commit might have caused this? If you tell me which one(s) you think is suspect can checkout a version prior to that and try it out now?

Ideally this sort of thing would be caught by an automated regression test, see #234.

tomwhite:
Fixes #235

This was inadvertently disabled in #209, and was the cause of #235 (I think), since it introduced lots of unecessary intermediate arrays. The plan for `lithops-matmul-random.py` goes from 3317 tasks to 2015 tasks with this change.

### DIFF
--- a/cubed/primitive/rechunk.py
+++ b/cubed/primitive/rechunk.py
@@ -1,8 +1,6 @@
 from math import ceil, prod
 from typing import Any, Dict, List, Optional, Tuple
 
-import zarr
-
 from cubed.primitive.types import CubedArrayProxy, CubedCopySpec, CubedPipeline
 from cubed.runtime.pipeline import spec_to_pipeline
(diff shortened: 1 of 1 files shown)


tomwhite:
I ran this successfully using `lithops-matmul-random.py`, so I think this is good to merge.

PR #252, opened 2023-07-07
Scaling docs

### LINKED ISSUE #204: Reasoning about scaling and performance
I tried to write out some thoughts about cubed's performance and scaling. Perhaps this could form the basis of another page of the docs, similar to how dask has pages on best practices and understanding performance. Obviously a lot of this will change as more optimizations are introduced / if we discover other reasons why scaling does not behave as expected.

## Aim of writing this out: 
- Understand better myself 
- Provide mental model for diagnosing performance 
- Explain to users what performance they should expect 
- Motivate [asv benchmarks](https://github.com/pydata/xarray/tree/main/asv_bench)
- Help us to ensure real scaling matches theoretical scaling 
- Material for docs page / future blog post 

## Types of scaling to explain:
- Horizontal versus vertical scaling 
- Weak scaling versus strong scaling 

### Theoretical scaling:
- Use rechunk as an example 
- Limited by concurrent writes to Zarr 
- Assuming serverless service provides infinite workers then...
- **Weak scaling should be totally linear**
- i.e. larger problem completes in same time given larger resources 

### Realistic scaling considerations:
- Actually requires parallelism, so won't happen with single-threaded executor 
- Weak scaling requires more workers than output chunks (might need to set executor config for big problems)
- Without enough workers strong scaling should be totally linear until more workers than output chunks 
- Stragglers will completely affect result - mitigate by turning on backups 
- Failures, once restarted, are basically stragglers
- Worker start-up 

## Scaling of a multi-step plan
- Number of steps in plan sets min total execution time 
- So reduce number of steps in plan 
- Array reductions can be done in fewer iterative steps if `allowed_mem` is larger (is there a rule of thumb for this??) (see https://github.com/tomwhite/cubed/issues/197#issuecomment-1576808572)
- Also can fuse steps 
- Happens automatically 
- Can't fuse through rechunking steps without requiring shuffle, which can violate memory constraint in general 
- But multiple blockwise operations can be fused (https://github.com/tomwhite/cubed/issues/136, https://github.com/tomwhite/cubed/issues/126)
- Can multiple rechunk operations be fused? 

## Multiple pipelines 
- i.e. two separate arrays you ask to compute simultaneously 
- Just requires enough workers for both to keep same weak scaling behaviour
- Same logic for two arrays which input into single array (or vice versa) 
- However currently won't compute in parallel on all executors (https://github.com/tomwhite/cubed/issues/80)

## Executor-specific considerations 
- Some worker startup time much faster than others 
- Different limits to max workers 
- If you used dask as an executor its scheduler might do unintuitive things - Does Beam have different properties?

TomNicholas:
Closes #204

Hopefully in the future this page might have less conjecture and more data :smile:

### DIFF
--- a/docs/user-guide/scaling.md
+++ b/docs/user-guide/scaling.md
@@ -1,92 +1,101 @@
-# Scaling and performance
+# Scaling
 
-Cubed is designed to scale to processing large arrays (~TB's) without losing performance.
+Cubed is specifically designed to maintain optimal performance while processing large arrays (of order Terabytes so far). 
+This page aims to provide a deeper understanding of how Cubed's design scales in both theoretical and practical scenarios.
 
-This page aims to help you understand how Cubed's design scales both in theory and in practice.
+## Preface: Types of Scaling
 
+There are different types of scaling to consider in distributed computing:
 
-## Preface: Types of scaling
+Horizontal versus Vertical Scaling: Horizontal scaling refers to adding more machines to the system to improve its throughput, while vertical scaling means enhancing the speed & resources of an existing machine.
 
-- Horizontal versus vertical scaling 
-- Weak scaling versus strong scaling
-    - Strong scaling is defined as how the solution time varies with the number of processors for a fixed total problem size.
-    - Weak scaling is defined as how the solution time varies with the number of processors for a fixed problem size per processor. 
+Weak versus Strong Scaling: Strong scaling is defined as how the solution time varies with the number of processors for a fixed total problem size. 
+Weak scaling is defined as how the solution time varies with the number of processors for a fixed problem size per processor. 
+In other words, strong scaling measures how much faster you can get a given problem done, whereas weak scaling measures how big a problem you can get done in a reasonable time.
 
-Cubed is designed to scale horizontally across a large number of machines in the cloud with the aim of providing good weak scaling properties when processing even very large array datasets.
+Cubed has been designed to scale horizontally, in such a way as to display good weak scaling even when processing large array datasets.
+
 
 ## Theoretical vs Practical Scaling of Cubed
 
+To understand factors affecting the scaling and performance of Cubed, we will start by considering the simplest possible calculation and add complexity.
+
 ### Single-step Calculation
 
-#### **Theoretical Scaling**
+The  simplest non-trivial operation in cubed would map one block from the input array to one block in the output array, with no intermediate persistent stores. 
+Changing the sign of every element in an array would be an example of this type of operation, known as an [elementwise](Operations/elemwise) operation.
 
-- Use rechunk as an example 
-- Limited by concurrent writes to Zarr 
-- Assuming serverless service provides infinite workers 
-- Weak scaling should be totally linear 
-- i.e. larger problem completes in same time given larger resources 
-- (Compute-bound step would have speedup limited by Ahmdahl's law??)
+In an ideal environment where the serverless service provides infinite workers, the limiting factor for scaling would be concurrent writes to Zarr. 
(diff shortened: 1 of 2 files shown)


docs/user-guide/scaling.md
```diff
@@ -0,0 +1,101 @@
+# Scaling
+
+Cubed is specifically designed to maintain optimal performance while processing large arrays (of order Terabytes so far).
+This page aims to provide a deeper understanding of how Cubed's design scales in both theoretical and practical scenarios.
+
+## Preface: Types of Scaling
+
+There are different types of scaling to consider in distributed computing:
+
+Horizontal versus Vertical Scaling: Horizontal scaling refers to adding more machines to the system to improve its throughput, while vertical scaling means enhancing the speed & resources of an existing machine.
```
tomwhite, line 10: "enhancing the speed & resources of an existing machine" - or replacing an existing machine with a larger one with more resources.

docs/user-guide/scaling.md
```diff
@@ -0,0 +1,101 @@
...
+Cubed has been designed to scale horizontally, in such a way as to display good weak scaling even when processing large array datasets.
+
+
+## Theoretical vs Practical Scaling of Cubed
+
+To understand factors affecting the scaling and performance of Cubed, we will start by considering the simplest possible calculation and add complexity.
+
+### Single-step Calculation
+
+The  simplest non-trivial operation in cubed would map one block from the input array to one block in the output array, with no intermediate persistent stores.
```
tomwhite, line 25: Nit: consistency of "Cubed" capitalisation.

docs/user-guide/scaling.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+
+## Other Performance Considerations
+
+### Different Executors
+
+Different executors come with their own set of advantages and disadvantages.
+For instance, some executors have much faster worker startup times, like Modal, while others may have different limits to the maximum workers.
+When using Dask as an executor, the scheduler might yield unintuitive results.
```
tomwhite, line 72: Can you explain why Dask might give unintuitive results, or give an example?

docs/user-guide/scaling.md
```diff
@@ -0,0 +1,101 @@
...
+### Different Cloud Providers
+
+Different cloud providers may perform differently. For example, Google Cloud Function (GCF) has been observed to have more stragglers than AWS.
+
+
+## Diagnosing Performance
+
+### Optimized Plan
+
+Use {py:meth}`Plan.visualize() <cubed.Plan.visualize()>` to view the optimized plan. This allows you to see the number of steps involved in your calculation.
```
tomwhite, line 84: ... and the number of tasks in each step, and overall.

tomwhite (approved):
LGTM!

docs/user-guide/scaling.md
TomNicholas: I'm not what I was thinking when I wrote this, I've removed it in favour of talking about there being differences between running on a cluster and running on serverless instead.

docs/user-guide/scaling.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+## Theoretical vs Practical Scaling of Cubed
+
+To understand factors affecting the scaling and performance of Cubed, we will start by considering the simplest possible calculation and add complexity.
+
+### Single-step Calculation
+
+The  simplest non-trivial operation in Cubed would map one block from the input array to one block in the output array, with no intermediate persistent stores.
+Changing the sign of every element in an array would be an example of this type of operation, known as an [elementwise](Operations/elemwise) operation.
```
TomNicholas, line 26: I'm not sure this link words - I've not used myst markdown for docs before

tomwhite (approved):
Thanks for the updates. I will merge (and fix links).

PR #269, opened 2023-07-20
Dask tests

### LINKED ISSUE #255: Unit tests for Dask executor
See https://github.com/tomwhite/cubed/pull/241#pullrequestreview-1520906419

tomwhite:
Fixes #255

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,6 +4,7 @@ omit =
     cubed/array_api/*
     cubed/extensions/*
     cubed/runtime/executors/beam.py
+    cubed/runtime/executors/dask.py
     cubed/runtime/executors/lithops.py
     cubed/runtime/executors/modal*.py
     cubed/vendor/*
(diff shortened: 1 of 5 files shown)


cubed/runtime/executors/dask.py
```diff
@@ -37,4 +42,15 @@ def execute_dag(
...
                     stage_delayed_funcs = [delayed_func]
 
-                dask.compute(*stage_delayed_funcs, **compute_kwargs)
+                def dask_callback(key, result, dsk, state, id):
+                    _, stats = result
+                    stats["array_name"] = name
+                    handle_callbacks(callbacks, stats)
+
+                if callbacks is not None:
+                    cm = Callback(posttask=dask_callback)
```
tomwhite, line 51: I just realised that this only works for local Dask schedulers, not for distributed. So probably best to drop this and exclude the callbacks test for Dask.

PR #278, opened 2023-07-27
Generalise async executor code

tomwhite:
Moves the logic for async executors into a common function, which can be used for an async Dask Distributed executor later.

Also adds a `batch_size` argument to address #239 for Modal (and Python async).


tomwhite:
The batching code needs more work (e.g. removing completed tasks from data structures), so I've removed it from this PR and will add separately later.

Issue #307, opened 2023-09-13
Move main array namespace


tomwhite:
Currently all the array functions are in `cubed.array_api`. This was created to follow the naming pattern for the new array API in `numpy.array_api`. Since then, however, there has been a change in emphasis to [view `numpy.array_api` as a reference implementation](https://github.com/data-apis/array-api-compat#difference-between-array_api_compat-and-numpyarray_api) so downstream libraries can test if they are using only array API functions - it is no longer recommended for end users to use it directly, since array libraries are slowly migrating their main namespaces to implement the array API. This includes [NumPy in version 2.0](https://hackmd.io/@seberg/Sk-a22Cmh?utm_source=preview-mode&utm_medium=rec#Add-array-API-standard-support-to-the-main-namespace), and in the meantime there is [array-api-compat](https://github.com/data-apis/array-api-compat) to ease migration for users of NumPy, CuPy, and PyTorch.

The upshot for Cubed is that we should probably move away from `cubed.array_api` for the above reasons, but also so we can add new functions that are not yet standardised, such as `nansum` (see #153).

The two options I see are:
1. `cubed`
2. `cubed.array`

TomNicholas:
> The two options I see are:
>
> 1. `cubed`
> 2. `cubed.array`
>
> Thoughts?

Either seems fine. (2) might be slightly better, because it would follow `dask.array`, and separate out the namespace for cubed things that aren't actually array operations (e.g. `Spec`).

tomwhite:
I went through the same thought process: `cubed.array` following `dask.array`, but then thinking that actually `cubed` is appropriate as Cubed is just an array library.

tomwhite:
Another very good explanation of the decision to move away from the separate `array_api` namespace pattern: https://github.com/numpy/numpy/issues/23883#issuecomment-1851429299

TomNicholas:
That's really interesting.

Related: I would like to test cubed(/pint/sparse/every other array library) by using [hypothesis.extra.array_api.make_strategies_namespace](https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.array_api.make_strategies_namespace), which requires passing in the array-API-compliant namespace directly. [See these docs](https://docs.xarray.dev/en/stable/user-guide/testing.html#compatibility-with-the-python-array-api-standard) (just released in the latest version of xarray) for an explanation of what I'm talking about.

Issue #153, opened 2023-03-14
np.nanmean executes eagerly on cubed arrays but lazily on dask arrays


TomNicholas:
See image for demonstration.

![Screenshot from 2023-03-14 19-26-13](https://user-images.githubusercontent.com/35968931/225164219-0125df14-e3f9-46ee-85c2-8ec523093ec1.png)

`np.nanmean` is called by xarray's `.mean()` method when `skipna=True`, which is the default.

tomwhite:
Dask supports the `__array_ufunc__` and `__array_function__` protocols which is why it works lazily in this case. Cubed doesn't, and `nanmean` is not in the Array API, so `np.nanmean` will coerce the Cubed array to a NumPy array eagerly.

There's a bit of a discussion about the nanfunctions in https://github.com/data-apis/array-api/issues/170, including  workarounds, but I'm not sure if that would help you @TomNicholas?

TomNicholas:
> nanmean is not in the Array API

Yeah I noticed that - what cubed does here is totally reasonable. I wasn't sure whether to raise this on cubed or xarray.

> There's a bit of a discussion about the nanfunctions in https://github.com/data-apis/array-api/issues/170,

Thanks! [This comment](https://github.com/data-apis/array-api/issues/170#issuecomment-827548927) mentions using the `where` kwarg to mean, which I presume is something we could change xarray to do instead of calling `nanmean` (though I expect `nanmean` is probably more performant?). The `where` kwarg also isn't in the Array API standard either though by the looks of it. They say there is another discussion about nan-aware functions but I can't find it by searching that repo :confused:

dcherian:
see https://github.com/pydata/xarray/pull/7424

PR #308, opened 2023-09-13
Add `nansum` and `nanmean` functions

tomwhite:
This is for #153 (although the Xarray side would need work too). I put the new functions in `cubed.array`, but there is a strong argument to put them (and all the existing array API functions) in the top-level `cubed` namespace (see #307).


cubed/core/ops.py
```diff
@@ -729,24 +729,20 @@ def reduction(
     allowed_mem = x.spec.allowed_mem
     max_mem = allowed_mem - x.spec.reserved_mem
 
-    # reduce initial chunks (if any axis chunksize is > 1)
-    if (
-        any(s > 1 for i, s in enumerate(result.chunksize) if i in axis)
-        or func != combine_func
-    ):
```
tomwhite, line 736: This condition needs to be removed since otherwise there are cases where no function is ever run in `reduction`. In particular if the array is of size 1 along the reduction axis. So far this hasn't been a problem since the reduction of a single element has always been that element (e.g. `sum(1)` is 1). But that doesn't hold for NaN (e.g. `nansum(nan)` is 0).

tomwhite:
> there is a strong argument to put them (and all the existing array API functions) in the top-level `cubed` namespace (see #307)

Now moved `nansum` and `nanmean` to `cubed`

Issue #187, opened 2023-05-26
Use google tensorstore to read/write to zarr?


TomNicholas:
I expect you're already aware of this @tomwhite , but I wanted to ask whether or not you thought the [google-tensorstore project](https://ai.googleblog.com/2022/09/tensorstore-for-high-performance.html) might be useful in cubed. @rabernat [suggested](https://discourse.pangeo.io/t/google-tensorstore-3d-data-package/2778) benchmarking its performance against zarr-python + fsspec. Presumably any improvement to the I/O speed would considerably increase the performance of cubed? Stephan recently released an [xarray interface to tensorstore](https://github.com/google/xarray-tensorstore) but it seems like that's not the right entry point for using it with cubed.

tomwhite:
Yes, I saw this and think it could be very interesting for Cubed.

> Presumably any improvement to the I/O speed would considerably increase the performance of cubed?

Hopefully, but we'd have to try it out to see.

tomwhite:
I tried swapping out Zarr for TensorStore here: https://github.com/tomwhite/cubed/tree/tensorstore. All unit test pass, except for `test_resume` (since TensorStore doesn't have a way of accessing Zarr's `nchunks_initialized`).

There were a few challenges due to differences in handling structured arrays, which we use for reductions like `mean`, but it looks like things would work. We probably will want a way to specify the array storage in the spec so users can select the one they want.

I haven't done any performance testing, but the current branch is sufficiently developed to try that.

PR #317, opened 2023-10-10
Array API backend

### LINKED ISSUE #315: Use array API for internal array operations
Cubed implements the [array API standard](https://data-apis.org/array-api/latest/), but it uses the regular NumPy API for its internal operations on chunks.

It should be possible to use the array API for these internal operations, which would then enable different array API libraries to be used in Cubed. JAX would be one possibility, for example, by using https://github.com/google/jax/pull/16099 (see also #304).

tomwhite:
Fixes #315

This uses https://github.com/data-apis/array-api-compat which makes NumPy conform to the array API. There are a couple of places where we use functions not in the array API (which fall back to NumPy): `take_along_axis` for arg reductions, and nan functions.

### DIFF
--- a/cubed/array_api/array_object.py
+++ b/cubed/array_api/array_object.py
@@ -19,6 +19,7 @@ from cubed.array_api.dtypes import (
     float64,
 )
 from cubed.array_api.linear_algebra_functions import matmul
+from cubed.backend_array_api import namespace as nxp
 from cubed.core.array import CoreArray
 from cubed.core.ops import elemwise
 from cubed.utils import memory_repr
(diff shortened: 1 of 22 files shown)


cubed/array_api/array_object.py
```diff
@@ -19,6 +19,7 @@
     float64,
 )
 from cubed.array_api.linear_algebra_functions import matmul
+from cubed.backend_array_api import namespace as nxp
```
alxmrs, line 22: Why not simply `import namespace as np`? This would minimize the diff.

cubed/backend_array_api.py
```diff
@@ -0,0 +1,23 @@
+import numpy as np
```
alxmrs, line 1: I'm excited to see this as simple place to drop in a Jax array implementation :) Great work.

cubed/array_api/array_object.py
tomwhite: It would, but I wanted to signal that this may not be regular NumPy. By using `nxp` here, it makes it easy to search for `np` to find places that are still using regular NumPy.

tomwhite:
Thanks for taking a look at this @alxmrs!

This has conflicts that I need to resolve, but it also conflicts with the work I've been doing for TensorStore in #187, which is why I opened #322. But maybe it's OK to merge this, and figure out the TensorStore part later?

cubed/backend_array_api.py
tomwhite: It will be interesting to see if this works with the [recent work in JAX on the Array API](https://github.com/google/jax/pull/16099).

TomNicholas:
> This uses https://github.com/data-apis/array-api-compat which makes NumPy conform to the array API.

You might be able to avoid the extra dependency on [array-api-compat](https://github.com/data-apis/array-api-compat) now that [numpy v1.26.0](https://github.com/numpy/numpy/releases/tag/v1.26.0) introduced `numpy.array_api`.

TomNicholas:
> But maybe it's OK to merge this, and figure out the TensorStore part later?

I vote merge this and think first about generalizing the array type in cases that don't also require changing the storage layer (e.g. cupy/sparse/pint).

tomwhite:
> > This uses https://github.com/data-apis/array-api-compat which makes NumPy conform to the array API.
> 
> You might be able to avoid the extra dependency on [array-api-compat](https://github.com/data-apis/array-api-compat) now that [numpy v1.26.0](https://github.com/numpy/numpy/releases/tag/v1.26.0) introduced `numpy.array_api`.

NumPy 1.22 introduced `numpy.array_api`, 1.26 updates it to the latest spec. The reason for using th compat layer is that it falls back to some functions that we use that are not in the standard yet ([like take_along_axis](https://github.com/tomwhite/cubed/pull/317/files#diff-9457623424e2d76236e4f0645ff382b6cc6ecf6f4daa0e43f5481d5961b403baR4-R7)).

If we're worried about the extra dependency then array-api-compat can be [vendored](https://github.com/data-apis/array-api-compat#vendoring), but I'm not sure that's necessary.

Issue #284, opened 2023-07-31
Fuse pipelines with different numbers of tasks


TomNicholas:
Last week @tomwhite and @dcherian and I discussed possible future optimizations for Cubed - this is my attempt to elucidate what I was suggesting.

### Motivation

The best-case scenario for a cubed computation is that all sequential operations get fused, because then no writing to intermediate stores is required. With no writing then every chunk moves through the whole calculation in parallel, despite multiple operations happening to it along the way. In general we can't guarantee zero intermediate stores being required because we also want to guarantee predictable memory usage during a full shuffle, but we might nevertheless aspire to fuse everything else :grin: 

### Idea

Currently Cubed's optimization pass fuses some blockwise operations together, but it can only fuse blockwise operations that have the [same number of tasks](https://github.com/tomwhite/cubed/blob/93ad984e7b0445164ab11b3c3f3a3b7db6c3bc97/cubed/primitive/blockwise.py#L251). If we could find a way to fuse blockwise operations with different numbers of tasks then potentially anything up to a full shuffle (see #282) could be fused.

### Use cases

It's possible to construct cubed plans in which blockwise operations with different numbers of tasks occur sequentially.

```python
from cubed.core.plan import visit_nodes

def print_num_tasks_per_pipeline(plan: cubed.core.Plan, optimize_graph: bool = False):
    """Print the number of tasks needed to execute each pipeline in this plan."""
    dag = plan.optimize().dag if optimize_graph else plan.dag.copy()
    for _, node in visit_nodes(dag, resume=None):
        print(f"{node['name']}: op_name = {node['op_name']}, num_tasks = {node['pipeline'].num_tasks}")
```

This can happen with `concat`:

```python
a = xp.asarray([[1, 2, 3], [4, 5, 6], [7, 8, 9]], chunks=(2, 2))
b = xp.asarray([[1, 1, 1], [1, 1, 1], [1, 1, 1]], chunks=(2, 2))
c = xp.concat([a + 1, b])
c.visualize()
```

![image](https://github.com/tomwhite/cubed/assets/35968931/016192cd-39e4-4a09-8e97-7be5f453ddb6)

(I don't really understand what all the side inputs are in these graphs - I hope they don't invalidate what I'm suggesting!)

```python
print_num_tasks_per_pipeline(c.plan)
```
```
array-010: op_name = blockwise, num_tasks = 4
array-013: op_name = blockwise, num_tasks = 6
```

Or `matmul`:

```python
import cubed.random

[...]

PR #327, opened 2023-11-21
Implementation of `std` and `var`.

### LINKED ISSUE #29: Implement `var` and `std`
Similar to the implementation of `mean`.

alxmrs:
Fixes #29. Here, I use existing cubed operations to implement `var` and `std`. Please let me know if I should reimplement the primitives as pure reductions.

### DIFF
--- a/cubed/array_api/__init__.py
+++ b/cubed/array_api/__init__.py
@@ -262,7 +262,7 @@ from .searching_functions import argmax, argmin, where
 
 __all__ += ["argmax", "argmin", "where"]
 
-from .statistical_functions import max, mean, min, prod, sum, std, var
+from .statistical_functions import max, mean, min, prod, std, sum, var
 
 __all__ += ["max", "mean", "min", "prod", "sum", "std", "var"]
 
--- a/cubed/array_api/__init__.py
+++ b/cubed/array_api/__init__.py
@@ -264,7 +264,7 @@ __all__ += ["argmax", "argmin", "where"]
 
 from .statistical_functions import max, mean, min, prod, std, sum, var
 
-__all__ += ["max", "mean", "min", "prod", "sum", "std", "var"]
+__all__ += ["max", "mean", "min", "prod", "std", "sum", "var"]
 
 from .utility_functions import all, any
 
--- a/cubed/array_api/__init__.py
+++ b/cubed/array_api/__init__.py
@@ -262,9 +262,9 @@ from .searching_functions import argmax, argmin, where
 
 __all__ += ["argmax", "argmin", "where"]
 
-from .statistical_functions import max, mean, min, prod, sum
+from .statistical_functions import max, mean, min, prod, sum, std, var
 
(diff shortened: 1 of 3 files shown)


cubed/array_api/__init__.py
```diff
@@ -244,9 +244,9 @@
 
 __all__ += ["argmax", "argmin", "where"]
 
-from .statistical_functions import max, mean, min, prod, sum
+from .statistical_functions import max, mean, min, prod, std, sum, var
 
-__all__ += ["max", "mean", "min", "prod", "sum"]
+__all__ += ["max", "mean", "min", "prod", "sum", "std", "var"]
```
alxmrs, line 249:
```suggestion
__all__ += ["max", "mean", "min", "prod", "std", "sum", "var"]
```

tomwhite:
Thank you for submitting a PR for this @alxmrs!

> Please let me know if I should reimplement the primitives as pure reductions.

That's how I envisaged doing this (similar to Dask). The way you have it at the moment has two separate reductions which will be less efficient than one (and reductions in Cubed are quite inefficient at the moment, something we want to remedy in e.g. #284). Also I'm not sure if there are any differences in numerical stability between the two approaches.

tomwhite:
It would be good to get std/var into Cubed. I found an implementation that we could try here (see https://github.com/cubed-dev/cubed/issues/29#issuecomment-2368108624) - would you be interested in having a go at that @alxmrs?

I wonder whether we should merge this in the meantime though?

cubed/array_api/statistical_functions.py
```diff
@@ -152,3 +153,12 @@ def sum(x, /, *, axis=None, dtype=None, keepdims=False):
         keepdims=keepdims,
         extra_func_kwargs=extra_func_kwargs,
     )
+
+
+def var(x, /, *, axis=None, keepdims=False):
```
tomwhite, line 158: This doesn't match the signature in the [specification](https://data-apis.org/array-api/latest/API_specification/statistical_functions.html) since it's missing a `correction` parameter. Similarly for `std`.

cubed/tests/test_array_api.py
```diff
@@ -563,6 +563,74 @@ def test_sum_axis_0(spec, executor):
     assert_array_equal(b.compute(executor=executor), np.array([12, 15, 18]))
 
 
+def test_var(spec, executor):
```
tomwhite, line 566: Can you parameterize all these tests for `var` for different axes, and for `keepdims`? There are other tests in the test suite that do this.

alxmrs:
Hey Tom. I don’t want to block you in the implementation of fundamental
routines. How about I try to get this merged in the next three days/EOW,
and if I can’t find the time, we close the PR? I appreciate how you’re
trying to incorporate my patches to cubed; thanks, I feel included.

tomwhite:
Thanks Alex - that sounds like a good plan!

(Sorry I didn't suggest how to move forward on this before - you weren't blocking me!)

tomwhite:
Fixed in #596

PR #329, opened 2023-11-21
Warn user when executing plan memory exceeds allowed memory.

### LINKED ISSUE #110: Issue warning if a computation actually used more memory than `allowed_mem`
We know the peak memory used in every task, so if this value ever exceeds the amount specified in `allowed_mem` (specified by the user) then we can issue a warning. We should encourage users to report these as bugs in cubed (though the immediate workaround would be to increase the memory available to the runtime).

alxmrs:
Fixes #110. I don't know if this is a good place to include this warning (during a call to `compute()`). Since array ops are lazily evaluated, there may not be a better place. Please let me know if you have ideas about where the warning could fit.

### DIFF
--- a/cubed/core/plan.py
+++ b/cubed/core/plan.py
@@ -2,6 +2,7 @@ import inspect
 import tempfile
 import uuid
 from datetime import datetime
+from warnings import warn
 
 import networkx as nx
 
@@ -165,6 +166,13 @@ class Plan:
         dag = self.optimize().dag if optimize_graph else self.dag.copy()
         dag = self.create_lazy_zarr_arrays(dag)
 
+        # TODO(#328): Consider caching the DAG somehow.
+        if self.max_projected_mem() > spec.allowed_mem:
+            warn(
+                "The projected maximum memory is above the allowed memory. This may be due to an error in Cubed. "
+                "Please report a bug: https://github.com/tomwhite/cubed/issues."
+            )
+
         if callbacks is not None:
             [callback.on_compute_start(dag, resume=resume) for callback in callbacks]
         executor.execute_dag(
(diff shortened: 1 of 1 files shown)


tomwhite:
Thanks for the PR @alxmrs!

The memory stuff is potentially confusing as there are quite a few different [definitions](https://tom-e-white.com/cubed/user-guide/memory.html) in Cubed! So I'll try to explain what I was trying to achieve here. As it stands, Cubed will already warn the user if the projected memory exceeds allowed memory *before* running the computation ([code](https://github.com/tomwhite/cubed/blob/d19caecdb148b46c89350123a6388adc39bff592/cubed/primitive/blockwise.py#L227-L230), [test](https://github.com/tomwhite/cubed/blob/d19caecdb148b46c89350123a6388adc39bff592/cubed/tests/test_core.py#L256-L260)).

What #110 is about is giving a warning if the *actual* memory usage exceeds allowed memory *after* the computation has run. This is referred to `peak_measured_mem` in https://tom-e-white.com/cubed/user-guide/memory.html.

The way it could be implemented would be to have a [callback](https://github.com/tomwhite/cubed/blob/d19caecdb148b46c89350123a6388adc39bff592/cubed/runtime/types.py#L33-L64) that collects all the `peak_measured_mem_end` values for all the tasks in a computation, then once the computation is over issues a warning if any exceed `allowed_mem`.

This is somewhat similar to the code in https://github.com/tomwhite/cubed/blob/d19caecdb148b46c89350123a6388adc39bff592/cubed/extensions/history.py, but the code there to analyze the memory usage is called manually by the user after the computation is over. The idea behind #110 is to warn the user without them having to configure anything.

tomwhite:
I've implemented this in #516

Issue #247, opened 2023-07-05
Don't materialize `full` or offsets arrays


tomwhite:
Arrays whose values are a function of the index do not need to be materialized - either in memory or on disk. These include arrays where every entry is the same (created with `full`), and arrays of offsets (used in `map_blocks` to provide block IDs to the map function).

Zarr arrays created with `full` are not expensive to create since they specify `write_empty_chunks=False` so no chunks are written - just the metadata. However, offsets are materialized to disk: https://github.com/tomwhite/cubed/blob/e21786591d9832a85f2e492641ad2061bdb8c14a/cubed/core/ops.py#L431-L432

Although the array is small since its size is the number of chunks, there is a scalability issue due to the fact that every chunk is stored in a separate file in the Zarr store.

These problems could be solved by writing an array implementation that implements indexing on the fly: either by returning a subarray of fill values for the case of `full`, or `np.ravel_multi_index` to turn an index into an offset for an array of offsets.

PR #336, opened 2023-12-17
Add VirtualInMemoryArray that keeps small arrays in memory

tomwhite:
This is a continuation of the idea in #247 and #290 - there is no need to materialize arrays to Zarr if they can be computed from the block ID (#247 and #290) or if they are small enough to be in memory anyway (this PR).

I've also added a check that the arrays created in this way (via `asarray`) are smaller than 1MB. Anything bigger is best written to shared storage (typically as a Zarr file), rather than be serialized in the plan (which is currently what is happening).

### DIFF
--- a/cubed/array_api/creation_functions.py
+++ b/cubed/array_api/creation_functions.py
@@ -7,9 +7,12 @@ from zarr.util import normalize_shape
 from cubed.backend_array_api import namespace as nxp
 from cubed.core import Plan, gensym
 from cubed.core.ops import map_direct
-from cubed.core.plan import new_temp_path
-from cubed.storage.virtual import virtual_empty, virtual_full, virtual_offsets
-from cubed.storage.zarr import lazy_from_array
+from cubed.storage.virtual import (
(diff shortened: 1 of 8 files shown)


dcherian:
> a check that the arrays created in this way (via asarray) are smaller than 1MB. 

Perhaps you can avoid this with the "broadcast trick" (https://github.com/dask/dask/issues/9517)
```
np.full(shape, 1) == np.ones(shape) == np.array([1]).broadcast_to(shape)
```

tomwhite:
That's a clever trick, thanks for pointing it out @dcherian! In this case I don't think it applies, since we have a small array of arbitrary values, not a constant array created by `full` - but it might be something we can use in that case.

tomwhite:
I just did a comparison of the number of Zarr arrays created during a unit test run:
* Main: 431 arrays
* This PR: 296 arrays

This probably isn't representative of real workloads since `asarray` isn't used much (and if it is, the arrays are small and don't dominate the computation).

Issue #69, opened 2022-07-20
Support multiple outputs in blockwise


tomwhite:
Currently we use Zarr structured arrays, which are likely to be slow (since they are not column based) - althought that would be worth checking first.

https://github.com/tomwhite/cubed/blob/400dc9adcf21c8b468fce9f24e8d4b8cb9ef2f11/cubed/array_api/statistical_functions.py#L18-L48

tomwhite:
Fixed in #419

PR #337, opened 2023-12-17
Change the DAG to have separate nodes for operations and arrays

tomwhite:
Currently there is a one-to-one correspondence between operations and arrays: one operation produces one array, and is represented by one node in the DAG.

However, this won't be true when we support multiple outputs (#69) where one operation can produce multiple arrays.

Similarly, it also won't support optimizations like sibling fusion (in [Apache Beam](https://beam.apache.org/releases/javadoc/2.0.0/org/apache/beam/sdk/transforms/ParDo.html), and from [FlumeJava](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/35650.pdf) originally), where operations that have the same inputs and produce different outputs can be fused into one operation producing multiple outputs. ([Wukong](https://github.com/ds2-lab/Wukong) calls this "task clustering".)

So we need to change the DAG representation so that one operation can produce multiple arrays, and this means breaking  nodes into two types: operations and arrays.

This PR does exactly that. An example DAG (from `test_add_with_broadcast`) in the current representation

cubed

now looks like this:

cubed-new-dag

Note:
* Operations are shown as boxes with rounded corners, arrays are rectangles.
* Colours indicate the following: pink for a blockwise operation, green for a rechunk operation, orange for a materialized array, white for an operation that is not run in its own pipeline or for an array that is not materialized to disk as a Zarr array ("virtual arrays")

I wondered about having an option to visualize DAGs as we currently do, but that won't work when there are multiple outputs, so it may be best just to have the more general representation.

### DIFF
--- a/cubed/core/plan.py
+++ b/cubed/core/plan.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from functools import lru_cache
 
 import networkx as nx
+import zarr
 
 from cubed.backend_array_api import backend_array_to_numpy_array
 from cubed.primitive.blockwise import can_fuse_pipelines, fuse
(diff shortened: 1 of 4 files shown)


tomwhite:
@TomNicholas any thoughts on this one before merging?

TomNicholas:
This looks great! I also asked my colorblind friend if he could distinguish all the colors and he gave it a thumbs up 👍

tomwhite:
> I also asked my colorblind friend if he could distinguish all the colors and he gave it a thumbs up 👍

Thank you!

Issue #331, opened 2023-11-27
Optimizing reduction


tomwhite:
The way that reduction currently works is illustrated here: https://tom-e-white.com/cubed/operations.html#reduction-and-arg-reduction. To summarize, each chunk in the input has a reduce function applied to it (with one task per chunk), the results are written to a new Zarr file, which then has `merge_chunks` called on it to combine into new chunks. This procedure is repeated until there is a single chunk in the reduction axis.

As suggested in #284, we could do much better by having one task process more than one chunk at once. This actually happens during subsequent rounds after the `merge_chunks` operation, but this is inefficient since the number of chunks that are merged is limited by the available memory since all the chunks are loaded into memory at once. If instead a task read chunks sequentially and called the reduce function before reading the next one, then the number of chunks processed by that task would not be limited by memory. This would also have the effect of making the number of chunks loaded by each task the same, which would make rounds of the reduction more uniform, and hence more predictable. In practice it would allow more chunks to be loaded per task, which would mean fewer rounds, and less data transferred. (It would be interesting to quantify this.)

This change would make the reduction algorithm more similar to tree reduce in Dask.

It would be worth comparing the effect of this change on performance with the existing reduction algorithm in Cubed for matrix multiplication, and Quadratic Means.

TomNicholas:
This sounds interesting :)

> As suggested in https://github.com/tomwhite/cubed/issues/284, we could do much better by having one task process more than one chunk at once. 
>
> If instead a task read chunks sequentially and called the reduce function before reading the next one, then the number of chunks processed by that task would not be limited by memory.

Would we be sacrificing some parallelism to do this?

TomNicholas:
Some notes from meeting with @tomwhite (and @saulshanabrook) today, in which I clarify my understanding:

1) **This suggestion is basically a `tree_reduce` for reductions**

We're suggesting that reductions be implemented using two distinct types of reduce step. Type (a) loads as many chunks as will fit in RAM, aggregates them, writes results to disk, then re-opens >1 of the new smaller chunks with a new container and repeats. The other type (b) opens more chunks than will fit in RAM, loads one at a time, sequentially aggregating each one as it is loaded. (Both of these have the bounded-memory property.)

[...]

PR #368, opened 2024-01-30
Fuse operations with different numbers of tasks

### LINKED ISSUE #284: Fuse pipelines with different numbers of tasks

tomwhite:
Fixes #284

This adds the necessary logic to the fuse code to fuse two operations that have a different number of tasks, but fusion must be controlled manually using the `always_fuse` argument to the optimization functions. This is fine for experimenting with the optimizer, but we'll probably want to add better heuristics about when to fuse these types of operations automatically in the future.

### DIFF
--- a/cubed/array_api/statistical_functions.py
+++ b/cubed/array_api/statistical_functions.py
@@ -25,7 +25,7 @@ def max(x, /, *, axis=None, keepdims=False):
     return reduction(x, nxp.max, axis=axis, dtype=x.dtype, keepdims=keepdims)
 
 
-def mean(x, /, *, axis=None, keepdims=False, use_new_impl=False):
+def mean(x, /, *, axis=None, keepdims=False, use_new_impl=False, split_every=None):
     if x.dtype not in _real_floating_dtypes:
         raise TypeError("Only real floating-point dtypes are allowed in mean")
     # This implementation uses NumPy and Zarr's structured arrays to store a
@@ -47,6 +47,7 @@ def mean(x, /, *, axis=None, keepdims=False, use_new_impl=False):
         dtype=dtype,
         keepdims=keepdims,
         use_new_impl=use_new_impl,
+        split_every=split_every,
         extra_func_kwargs=extra_func_kwargs,
     )
 
--- a/cubed/core/ops.py
+++ b/cubed/core/ops.py
@@ -27,10 +27,16 @@ from cubed.core.plan import Plan, new_temp_path
 from cubed.primitive.blockwise import blockwise as primitive_blockwise
 from cubed.primitive.blockwise import general_blockwise as primitive_general_blockwise
 from cubed.primitive.rechunk import rechunk as primitive_rechunk
-from cubed.utils import chunk_memory, get_item, offset_to_block_id, to_chunksize
+from cubed.utils import (
+    _concatenate2,
+    chunk_memory,
+    get_item,
+    offset_to_block_id,
+    to_chunksize,
+)
 from cubed.vendor.dask.array.core import common_blockdim, normalize_chunks
 from cubed.vendor.dask.array.utils import validate_axis
-from cubed.vendor.dask.blockwise import broadcast_dimensions
+from cubed.vendor.dask.blockwise import broadcast_dimensions, lol_product
 from cubed.vendor.dask.utils import has_keyword
 
 if TYPE_CHECKING:
@@ -764,6 +770,7 @@ def rechunk(x, chunks, target_store=None):
 
 
 def merge_chunks(x, chunks):
+    """Merge multiple chunks into one."""
     target_chunksize = chunks
     if len(target_chunksize) != x.ndim:
         raise ValueError(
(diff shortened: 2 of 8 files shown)


cubed/array_api/statistical_functions.py
```diff
@@ -47,6 +47,7 @@ def mean(x, /, *, axis=None, keepdims=False, use_new_impl=False):
         dtype=dtype,
         keepdims=keepdims,
         use_new_impl=use_new_impl,
+        split_every=split_every,
```
TomNicholas, line 50: Is this `split_every` argument a temporary implementation detail whilst we figure out fusing heuristics? It would be nice to relate the meaning of this argument back to the discussion in #284 (as I currently don't quite understand what it means)

TomNicholas:
This is very exciting!

cubed/array_api/statistical_functions.py
tomwhite: The `split_every` argument is what we referred to as "fan-in" in #284 - the number of chunks read by one task doing the reduction step. It's the same in Dask and can be a dictionary indicating the number of chunks to read in each dimension.

I hope it is something that we can get better heuristics for (or at least good defaults) - possibly by measuring trade offs like they did in the Primula paper, see #331.

tomwhite:
I've generated a few plan visualizations to give an idea of what the code does.

Running this cut-down quadratic means example

https://github.com/tomwhite/cubed/blob/1762d3c685e7a1333c2809d2d9e419ff6508c5ce/cubed/tests/test_core.py#L527-L537

gives the following plan using the old optimization algorithm (for `t_length=500`):

quad_means_current

The reduce rounds (for the `mean` operations) have 50, 4, and 1 task(s).

If we use the new optimization algorithm then it will fuse the first `mean` operation that has two inputs into one:

```python
    m = xp.mean(uv, axis=0, split_every=10, use_new_impl=True)

    m.visualize(
        filename=tmp_path / "quad_means",
        optimize_function=multiple_inputs_optimize_dag
    )

```
quad_means_fuse_multiple_inputs

Notice that the number of tasks has gone down from 160 to 59, and the amount of intermediate data stored has gone down from 16.8GB to 7.7GB. The reduce rounds have 50, 5, and 1 task(s) - this is controlled by the `split_every=10` argument.

If we now fuse the first two operations (purple boxes with rounded corners) then we get make the plan even smaller. Note in the code below we have to explicitly ask for this operation to be fused.

This is using the logic from this PR which is able to fuse operations with a different number of tasks. The first operation (marked `__mul__`) has 50 tasks, and the second (`mean`) has 5.

```python
    m = xp.mean(uv, axis=0, split_every=10, use_new_impl=True)

    m.visualize(
        filename=tmp_path / "quad_means",
        optimize_function=partial(multiple_inputs_optimize_dag, always_fuse=["op-008"]),
    )
```

quad_means_fuse_diff_num_tasks

There are now only two reduce rounds: with 5 and 1 task(s). And a total of only 8 tasks, and 166.8MB of intermediate data - a significant saving.

tomwhite:
The mypy failures suggested to me that `num_input_blocks` belongs to `BlockwiseSpec`, not `PrimitiveOperation`, since it doesn't apply to a `rechunk` operation, for example. This change means that `num_input_blocks` also sits next to `function_nargs`, which it is related to, so that seems better.

TomNicholas:
Wow that's a big difference. Thanks for writing this out @tomwhite.

I was about to try an estimate the projected speedup here (assuming IO dominates the execution time), but without being able to see the number of tasks in each stage it's hard to reason about. (Half the number of zarr stores to be written, but not necessarily twice as fast because do `array-008` and `array-007` have the same amount of parallelism that they did before optimization?) Not seeing this information in the visualization also made https://github.com/coiled/feedback/issues/271#issuecomment-1921663860 less clear - perhaps this information should be included in the graph visualization itself?

tomwhite:
It's definitely hard to reason about, even with the number of tasks in each stage. (BTW I have listed the number of tasks in each stage in the comment above.) The total number of stages is probably the biggest factor in how long the computation takes. I plan to run some benchmarks to measure the performance of these workloads.

But I agree that it would be useful to put the number of tasks in the visualization. It does appear in the tooltip for nodes, but for some reason that doesn't appear when the SVG is embedded in another page.

PR #369, opened 2024-02-01
small changes to coiled/aws examples

rsignell:
Just a few minor changes that were necessary for me to run the examples on coiled.   I needed to run with the latest version of coiled (I did `pip install coiled==1.3.12.dev12`), for more info see https://github.com/coiled/feedback/issues/271

### DIFF
--- a/examples/coiled/aws/coiled-add-asarray.py
+++ b/examples/coiled/aws/coiled-add-asarray.py
@@ -21,7 +21,7 @@ if __name__ == "__main__":
     c = xp.add(a, b)
     res = c.compute(
         executor=executor,
-        memory=["1 GiB", "8 GiB"] # memory range
+        memory=["1 GiB", "8 GiB"], # memory range, lower value must be at least allowed_mem
         spot_policy="spot_with_fallback",  # recommended
         account=None,  # use your default account (or change to use a specific account)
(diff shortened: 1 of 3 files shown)


examples/coiled/aws/coiled-add-asarray.py
```diff
@@ -21,7 +21,7 @@
     c = xp.add(a, b)
     res = c.compute(
         executor=executor,
-        memory="1 GiB",  # must be greater than allowed_mem
+        memory=["1 GiB", "8 GiB"] # memory range
```
tomwhite, line 24: Looks like there's a missing comma here?

Also might be worth preserving the previous comment too:

```
# memory range, lower value must be at least allowed_mem"
```

tomwhite (commented):
Thanks for submitting this @rsignell!

examples/coiled/aws/coiled-add-asarray.py
TomNicholas: I didn't realise we could submit a range now, that's nice

tomwhite: To be clear - the `memory` argument is passed through to the Coiled function, it's not interpreted by Cubed.

tomwhite:
Thanks for the contribution @rsignell

PR #371, opened 2024-02-05
Show number of tasks prominently on graph visualization

### LINKED ISSUE #370: Include number of tasks in operation nodes in graph visualization
It's in the tooltips, but these don't always appear, see https://github.com/tomwhite/cubed/pull/368#issuecomment-1921695717

tomwhite:
Fixes #370 

Here's an example:

cubed-add


TomNicholas:
Nice! Yeah this gives much more idea of the scale of what will happen at each point in the graph.

Issue #304, opened 2023-09-10
Jax integration


alxmrs:
Can the core array API ops of cubed be implemented in jax, s.t. everything easily compiles to accelerators? Could this solve the common pain point of running out of GPU memory? How would other constraints (GPU bandwidth limits) be handled? What is the ideal distributed runtime environment to make the most of this? Could spot GPU instances be used (serverless accelerators)?

tomwhite:
Thanks @alxmrs for opening this issue. I'm not familiar enough with JAX or GPUs to answer these questions, but I'd be happy to support or discuss an initiative in this direction. Is there a small piece of work that you have in mind that could be used to explore this?

alxmrs:
The Jax docs may provide a few good toy examples useful to validate this idea. 

Check out this tutorial: https://jax.readthedocs.io/en/latest/jax-101/06-parallelism.html

This tutorial on distributing computation on a pool of TPUs (specifically, the neural net section) may be of interest, too:

https://jax.readthedocs.io/en/latest/notebooks/Distributed_arrays_and_automatic_parallelization.html#examples-neural-networks

High level goals for Jax + Cubed may be to make managing GPU memory effortless: 
- https://jax.readthedocs.io/en/latest/gpu_memory_allocation.html#common-causes-of-oom-failures
- E.g.: could we do an FFT on a 100 GiB (TiB? PiB??) array? https://github.com/google/jax/discussions/13842

tomwhite:
Thanks for the pointers @alxmrs.

Thinking about how Cubed might hook into this, the main idea in Cubed is that every array is backed by Zarr, and an operation maps one Zarr array to another, by working in parallel on chunks (see https://tom-e-white.com/cubed/cubed.slides.html#/1).

Every operation is ultimately [expressible as a blockwise operation](https://tom-e-white.com/cubed/operations.html) (or a rechunk operation, but let's ignore that here), which ends up in the `apply_blockwise` function:

https://github.com/tomwhite/cubed/blob/0d13e4f2b12c22d1c41b9f4ea693266b21d808d0/cubed/primitive/blockwise.py#L53-L75

The key parts are:
1. Reading each arg from Zarr into (CPU) memory (line 67),
2. Invoking the function on the args (line 70), and
3. Writing the result from (CPU) memory back to Zarr (line 73 or 75)

[...]

PR #374, opened 2024-02-09
Run tests using JAX as the backend array API (on CPU)

tomwhite:
This is to show that things work when the backend array library is something other than NumPy.

### DIFF
--- a/cubed/storage/virtual.py
+++ b/cubed/storage/virtual.py
@@ -5,6 +5,7 @@ import numpy as np
 import zarr
 from zarr.indexing import BasicIndexer, is_slice
 
+from cubed.backend_array_api import backend_array_to_numpy_array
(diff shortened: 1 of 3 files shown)


tomwhite:
This is related to #304

cubed/storage/virtual.py
```diff
@@ -129,7 +130,7 @@ def __init__(
         self.chunks = template.chunks
         self.template = template
         if array.size > 0:
-            template[...] = array
+            template[...] = backend_array_to_numpy_array(array)
```
TomNicholas, line 133: FYI xarray has some more general versions of these types of array coercion functions here

https://github.com/pydata/xarray/blob/f343b7b23a275ebb17cdd91e73289525a627d39e/xarray/core/pycompat.py#L109

PR #378, opened 2024-02-12
Enable backup tasks by default for all executors that support them

tomwhite:
The policy for launching backup tasks for stragglers is fairly conservative, and I haven't seen any problems while using it. In fact it has occasionally kicked in to help a slow-running computation, so I think it would be good to enable it by default.


TomNicholas:
I agree, when I ran those quadratic means cases a while ago it seemed to only ever improve things.

PR #395, opened 2024-02-22
Create a single set of example scripts that can run on any executor

tomwhite:
This is not ready for merging, but shows how the examples can be simplified so they can run on any executor.

From the _examples_ directory, run as follows:

```shell
export LITHOPS_CONFIG_FILE=~/.lithops/config.aws
CUBED_CONFIG=$(pwd)/lithops/aws-lambda python add-random.py
```


TomNicholas:
Cool!

tomwhite:
A similar approach should work for benchmarking too.

tomwhite:
I've updated and tested Lithops, Modal, and Coiled with the new changes. (I've left Beam unchanged as it uses a command line parser for its arguments.)

This is not ready to be merged yet, as it relies on changes that have not been released. I'd like to merge #396, #397, and #398 and then do a release in a day or so. @TomNicholas please let me know if there's anything else you'd like in the release.

TomNicholas:
Looks great! No requests.

Issue #381, opened 2024-02-15
Document optimization changes


tomwhite:
Update the documentation to reflect the changes due to the work in #339. This should include end user docs on how to configure/control optimization, as well as more developer-focused design docs on the implementation.

tomwhite:
Fixed in #397

PR #397, opened 2024-02-26
Add page on optimization to user guide

tomwhite:
See #381.

This relies on #396 to make it easier to cross-refer operation IDs in the debug output to the visualizations.


docs/user-guide/optimization.md
```diff
@@ -0,0 +1,109 @@
+# Optimization
+
+Cubed will automatically optimize the computation graph before running it. This can reduce the number of tasks in the plan, and the amount of intermediate IO, both of which speed up the computation.
```
TomNicholas, line 3: From this intro it is unclear why I would ever not want to fuse tasks. Are there situations in which fusing would sacrifice memory safety? Or is it merely a trade-off between asking single tasks to read more data and potentially having more IO steps in the pipeline?

docs/user-guide/optimization.md
```diff
@@ -0,0 +1,109 @@
...
+
+Now we call `visualize` again, this time not setting `optimize_graph` so it picks up its default value of `True`.
+
+```python
+c.visualize()
+```
+
+![Map fusion optimization](../images/optimization_map_fusion.svg)
+
+Note that with optimization turned on, the array `b` is no longer written as an intermediate output since it will be computed in the same tasks that compute array `c`. The overall number of tasks is reduced from 10 to 5, and the intermediate data (total nbytes) is reduced too.
```
TomNicholas, line 31: Maybe note that this optimization is always beneficial, which is why it's on by default?

docs/user-guide/optimization.md
```diff
@@ -0,0 +1,109 @@
...
+To change this setting, you can use Python's `functools.partial` as follows:
+
+```python
+opt_fn = partial(multiple_inputs_optimize_dag, max_total_source_arrays=8)
+visualize(optimize_function=opt_fn)
+```
+
+`max_total_num_input_blocks` specifies the maximum number of input blocks (chunks) that are allowed in the fused operation.
+
+Again, this is to limit the number of reads that an individual task must perform. The default is `None`, which means that operations are fused only if they have the same number of tasks. If set to an integer, then this limitation is removed, and tasks with a different number of tasks will be fused - as long as the total number of input blocks does not exceed the maximum. This setting is useful for reductions, and can be set using `functools.partial`:
```
TomNicholas, line 80: My understanding is that both the `max_total_source_arrays` and `max_total_num_input_blocks` are related to the "optimal fan-in" parameter described in #331. I therefore think this explanation would be better motivated if it started with a description of that general trade-off first. 

I would have structured it like this:
- Description of the trade-off detailed in #331
- Therefore an optimal fan-in parameter exists (unique for any particular reduction?)
- Cubed makes a default global choice for you, but you may want to override it
- In which scenarios are you likely to want to override this default?
- Here's how you actually use the API to change the default...

docs/user-guide/optimization.md
```diff
@@ -0,0 +1,109 @@
...
+
+logging.basicConfig(level=logging.DEBUG)
+
+e.visualize(optimize_function=multiple_inputs_optimize_dag)
+```
+
+The output explains which operations can or can't be fused, and why:
+
+```
+DEBUG:cubed.core.optimization:can't fuse op-001 since it is not fusable
```
TomNicholas, line 104: that's not the most helpful message 😆 - why is it not fusable?

tomwhite:
> This is very helpful for me to understand what optimization options we have, but for a user this would seem like an unmotivated exposure of pretty deep internals. Is this intended as developer docs?

I wrote it to capture what I'd implemented so far on the optimization work, but you're right that it's not pitched well for users!

I hope that we can make the more aggressive optimizations enabled by default in which case users won't need to care about these settings, although there will likely always be cases where an advanced user wants to change things. But we're not ready to do that quite yet. 

Might be worth leaving this open until we have more experience with optimizing some example workloads.

docs/user-guide/optimization.md
tomwhite: > Are there situations in which fusing would sacrifice memory safety?

No, fusing should always respect memory limits.

> Or is it merely a trade-off between asking single tasks to read more data and potentially having more IO steps in the pipeline?

Yes that's it.

docs/user-guide/optimization.md
tomwhite: Oops - that is a tautology! In this case, `op-001` has no predecessors so there's nothing to fuse with.

tomwhite:
> To be clear - as a temporary record of the state of possible optimizations, this is great. My criticisms are more looking towards what would be most useful for users in the future.

It's very helpful feedback - thanks!

tomwhite:
I've re-worked this page to reflect the new defaults (multiple-input fusion is enabled by default now), and also to move the advanced settings to the end since they should be rarely needed. I think this is generally useful for users should should be Ok to go in now @TomNicholas.

docs/user-guide/optimization.md
```diff
@@ -0,0 +1,120 @@
...
+DEBUG:cubed.core.optimization:can't fuse op-003 since it is not a primitive operation, or it uses map_direct
+DEBUG:cubed.core.optimization:can't fuse op-004 since no predecessor ops can be fused
+DEBUG:cubed.primitive.blockwise:can fuse op-005 since num tasks of predecessor ops match
+```
+
+## Advanced settings
+
+There are limits to how many input arrays and input chunks reads are fused together. These are imposed so that the number of reads that an individual task must perform is not excessive, which would otherwise result in slow running tasks.
+
+In some cases you may want to change these limits, which we look at here.
```
TomNicholas, line 95: Do we want to explain what some of the cases might be? Or leave that as future work?

tomwhite: It could certainly be expanded, but I'd prefer to leave it for now.

TomNicholas (approved):

PR #402, opened 2024-02-29
Bring vecdot implementation in line with the one in array-api-compat

tomwhite:
This fixes `array_api_tests/test_linalg.py::test_vecdot` that was failing.


tomwhite:
This is failing with a different error in the tensordot test

tomwhite:
This should be re-worked after #481 is in

tomwhite:
This is failing for some edge cases for arrays of size 0 (tracked in #420), so I haven't re-enabled array API tests for this.

PR #412, opened 2024-03-05
Make the new multiple-inputs optimization function the default.

tomwhite:
I've been testing this a bit at scale, and I think it's ready to enable the new optimization function (`multiple_inputs_optimize_dag`) as the default. (#397 explains how it works in more detail.)


tomwhite:
The `test_expand_dims` failure in the array API tests has found a real bug which I'm looking into.

tomwhite:
This also enables the new reduction implementation by default. It's possible to disable on a per-function basis if needed by calling e.g. `xp.mean(x, use_new_impl=False)`.

Issue #432, opened 2024-03-19
Providing filesystem credentials to cubed.Spec


balanz24:
Currently cubed doesn't allow to specify the credentials of a storage file system to `Spec` class. Although the class allows to pass some `storage_options`, these are actually never used.

This is a limitation since when working with S3-compatible storage system such as minIO the user should be able to specify credentials.

The solution could come from:

1. Specifying credentials and addtional configuration parameters to Spec:
```python
spec = cubed.Spec(work_dir="s3://cubed", allowed_mem='2GB', 
reserved_mem='100MB', executor_name="lithops", 
storage_options={"client_kwargs":{"endpoint_url": 'http://<ip>:<port>'}, "key": "key", "secret": "secret"})
```

2. Wrapping the store path with specs storage_options when creating the temp_store and the target_store. e.g. in [rechunk](https://github.com/cubed-dev/cubed/blob/e185fca6d12a9a2d95a82ec24be69e34cbfcefad/cubed/core/ops.py#L739)
```python
if target_store is None:
        target_store = new_temp_path(name=name, spec=spec) # return not only path, but also store arguments
name_int = f"{name}-int"
temp_store = new_temp_path(name=name_int, spec=spec) # return not only path, but also store arguments
```

3. Before calling [zarr.open_array](https://github.com/cubed-dev/cubed/blob/e185fca6d12a9a2d95a82ec24be69e34cbfcefad/cubed/storage/zarr.py#L51), checking if `self.store` is just a string or a string packed with additional arguments. If that is the case create a S3 compatible storage class as:
```python

if isintance(self.store, dict):
  s3 = s3fs.S3FileSystem(**self.store["storage_options"])
  self.store = s3fs.S3Map(root=self.store["path"], s3=s3, check=False)

target = zarr.open_array(
            self.store,
            mode=mode,
            shape=self.shape,
            dtype=self.dtype,
            chunks=self.chunks,
            path=self.path,
            fill_value=self.fill_value,
            **self.kwargs,
        )
```

As accessing a MinIO server from zarr, for instance, is done by:

```python
s3 = s3fs.S3FileSystem(key='key', secret='secret', client_kwargs={"endpoint_url": 'http://<ip>:<port>' })
store = s3fs.S3Map(root='s3://cubed/test.zarr', s3=s3, check=False)
a = zarr.open_array(shape=(10, 100, 100), chunks = (1, 100, 100), dtype = 'float64', store=store)
````

tomwhite:
[...]

PR #436, opened 2024-03-22
Providing filesystem credentials through storage_options

balanz24:
Fixes [#432](https://github.com/cubed-dev/cubed/issues/432#issue-2194715891). This allows to specify the credentials of a storage file system to Spec class. Needed for S3-compatible storage system such as minIO.

### DIFF
--- a/cubed/core/ops.py
+++ b/cubed/core/ops.py
@@ -290,6 +290,7 @@ def blockwise(
     spec = check_array_specs(arrays)
     if target_store is None:
         target_store = new_temp_path(name=name, spec=spec)
+    
     op = primitive_blockwise(
         func,
         out_ind,
@@ -308,6 +309,7 @@ def blockwise(
         extra_func_kwargs=extra_func_kwargs,
         fusable=fusable,
         num_input_blocks=num_input_blocks,
+        storage_options=spec.storage_options,
         **kwargs,
     )
     plan = Plan._new(
@@ -367,6 +369,7 @@ def general_blockwise(
         in_names=in_names,
         extra_func_kwargs=extra_func_kwargs,
         num_input_blocks=num_input_blocks,
+        storage_options=spec.storage_options,
         **kwargs,
     )
     plan = Plan._new(
@@ -746,6 +749,7 @@ def rechunk(x, chunks, target_store=None):
         reserved_mem=spec.reserved_mem,
         target_store=target_store,
         temp_store=temp_store,
+        storage_options=spec.storage_options,
     )
 
     from cubed.array_api import Array
--- a/cubed/core/ops.py
+++ b/cubed/core/ops.py
@@ -290,7 +290,6 @@ def blockwise(
     spec = check_array_specs(arrays)
     if target_store is None:
         target_store = new_temp_path(name=name, spec=spec)
-    
     op = primitive_blockwise(
         func,
         out_ind,
--- a/cubed/storage/zarr.py
+++ b/cubed/storage/zarr.py
@@ -1,6 +1,7 @@
 from typing import Optional, Union
 
 import zarr
+import s3fs
 
 from cubed.types import T_DType, T_RegularChunks, T_Shape, T_Store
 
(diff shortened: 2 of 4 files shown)


cubed/core/ops.py
```diff
@@ -308,6 +308,7 @@ def blockwise(
         extra_func_kwargs=extra_func_kwargs,
         fusable=fusable,
         num_input_blocks=num_input_blocks,
+        storage_options=spec.storage_options,
```
tomwhite, line 311: I know they are all kwargs, but the `storage_options` parameter would be better immediately after `target_path`, since it belongs with `target_store` and `target_path`. This applies to all the changes in ops.py and blockwise.py, but not rechunk.py.

cubed/storage/zarr.py
```diff
@@ -20,18 +21,21 @@ def __init__(
...
         **kwargs,
     ):
         """Create a Zarr array lazily in memory."""
         # use an empty in-memory Zarr array as a template since it normalizes its properties
         template = zarr.empty(
             shape, dtype=dtype, chunks=chunks, store=zarr.storage.MemoryStore()
         )
+        if storage_options:
+            s3 = s3fs.S3FileSystem(**storage_options)
+            store = s3fs.S3Map(root=store, s3=s3, check=False)
```
tomwhite, line 34: This code is S3-specific, which we don't want to hardcode. In fact, I think you don't need any changes in zarr.py at all - since the `storage_options` from the spec will be passed straight through to [`zarr.open_array`](https://zarr.readthedocs.io/en/stable/api/creation.html#zarr.creation.open_array). Does that work in your case?

tomwhite (commented):
This looks great - thanks for submitting it @balanz24.

I've made a couple of review comments, and some formatting fixes are needed too. You can install `pip install pre-commit` then run `pre-commit run --all-files` to fix them.

cubed/storage/zarr.py
balanz24: You are right. They can be passed through **kwargs. 
However kwargs were not passed to `zarr.open_array` inside `LazyZarrArray.open()` so I made that modification.

balanz24:
Thanks for the comments @tomwhite.

I was looking at [`cubed.from_zarr`](https://github.com/cubed-dev/cubed/blob/main/cubed/core/ops.py#L94) function which also makes a call to `zarr` API. Maybe it should also receive **kwargs in order to allow `storage_options`  be passed?

tomwhite:
> Or extract them from `spec` which is already passed?

This is the most consistent way of doing it.

tomwhite:
It would be good to add a unit test. Perhaps use a fsspec local filesystem and set `auto_mkdir` to a non-default value (i.e. to `True`) to check that it has been passed through. Could this work?

cubed/core/ops.py
```diff
@@ -109,7 +109,7 @@ def from_zarr(store, path=None, spec=None) -> "Array":
         The array loaded from Zarr storage.
     """
     name = gensym()
-    target = zarr.open(store, path=path, mode="r")
+    target = zarr.open(store, path=path, mode="r", storage_options=spec.storage_options)
```
tomwhite, line 112: This will fail if `spec` is `None`, which it can be when the default config is being used. You can fix this by adding a line like this:

https://github.com/cubed-dev/cubed/blob/3d08513359ef168979a372a702d2f41f79e6f769/cubed/core/array.py#L47

balanz24: Done. Sorry for the confussion.

balanz24:
Regarding the unit test, I'm trying something like this:
```python
def test_storage_options(tmp_path):
    zarr_path = f"file://{tmp_path}/dir/array.zarr"
    arr = lazy_zarr_array(zarr_path, shape=(3, 3), dtype=int, chunks=(2, 2), storage_options={"auto_mkdir": False})

    with pytest.raises(ValueError):
        arr.open()

    arr.create()
```
The problem is that the non-existing intermediate directory gets created with `auto_mkdir` set to `False`.

tomwhite:
> The problem is that when running it locally, the non-existing intermediate directory gets created with `auto_mkdir` set to `False`.

Ah, it looks like Zarr is doing this:

https://github.com/zarr-developers/zarr-python/blob/37e0a1a0c22f552daf6fd94fec5474a9b92db33d/zarr/storage.py#L1376

So that's not going to work as a test. I can't think of another simple way of testing this - have you got any ideas? Also, how has the manual testing been going?

balanz24:
With this current version the credentials are passed fine and I'm able to use a minIO server as the storage backend.

I don't know how can we test it without having access to an external filesystem. It seems that in linux you can't create a directory with credentials through POSIX, nor simulate an S3 API over local directories.

tomwhite:
I'm going to merge this now. I've opened #439 for adding a test.

Thank you for your contribution @balanz24!

PR #442, opened 2024-04-11
Group by

tomwhite:
Adds a `groupby_reduction` function for doing "grouped reductions". This is intended for [flox](https://flox.readthedocs.io/en/latest/) to add Cubed support.

The only change needed to the regular `reduction` function is the addition of a `combine_sizes` parameter, which makes it possible to specify the size of an axis after the combine function has been applied. In normal reductions the reduction axes are always 1 (the `combine_sizes` default), but in a grouped reduction the axis representing the groups is instead the number of groups.


cubed/core/groupby.py
```diff
@@ -0,0 +1,89 @@
...
+if TYPE_CHECKING:
+    from cubed.array_api.array_object import Array
+
+
+def groupby_reduction(
+    x: "Array",
+    by: "Array",
+    func,
+    combine_func=None,
+    aggegrate_func=None,
```
TomNicholas, line 16: typo

```suggestion
    aggregate_func=None,
```

tomwhite: Well spotted! See #443

PR #453, opened 2024-04-29
Test on Python 3.12

tomwhite:


tomwhite:
This is failing because Apache Beam doesn't support Python 3.12 yet (https://github.com/apache/beam/issues/29149). It probably makes sense to split out the beam tests into another GHA workflow file.

PR #473, opened 2024-06-04
Top level array namespace

tomwhite:
This is a part of #307, which simply adds the array API public members to the top-level `cubed` namespace. Later (probably after the next release), I'd like to deprecate everything in `cubed.array_api` for removal in the future.

Any thoughts @TomNicholas?


TomNicholas:
This seems sensible, my main question is just whether this would prevent it working with hypothesis for any reason (see https://github.com/cubed-dev/cubed/issues/307#issuecomment-1854571104)

tomwhite:
> This seems sensible, my main question is just whether this would prevent it working with hypothesis for any reason (see [#307 (comment)](https://github.com/cubed-dev/cubed/issues/307#issuecomment-1854571104))

Seems to work fine:

```python
>>> from hypothesis.extra.array_api import make_strategies_namespace
>>> import cubed as xp
>>> xps = make_strategies_namespace(xp)
>>> xps.arrays(xp.int8, (2, 3)).example()
cubed.Array<array-169, shape=(2, 3), dtype=int8, chunks=((2,), (3,))>
>>> import cubed.array_api as xp
>>> xps = make_strategies_namespace(xp)
>>> xps.arrays(xp.int8, (2, 3)).example()
cubed.Array<array-2396, shape=(2, 3), dtype=int8, chunks=((2,), (3,))>
```

PR #478, opened 2024-06-10
Speed up tests

JosepSampe:
Based on the discussion [here](https://github.com/cubed-dev/cubed/pull/456):
- Disable the Lithops `module_manager` to speed up tests in all the OS
- Speed up tests on python3.12 by setting `COVERAGE_CORE: sysmon`


tomwhite (approved):
That's a lot faster - thanks @JosepSampe!

Issue #295, opened 2023-08-22
Test with Zarr V3


tomwhite:
Add a GH Actions workflow that runs Cubed tests using Zarr V3 storage. This will ensure that the upcoming work to update Zarr to support the V3 spec (see https://github.com/zarr-developers/zarr-python/discussions/1480) is working in Cubed.

I've done a preliminary test in https://github.com/tomwhite/cubed/tree/zarr-v3-testing, and the current V3 implementation works with Cubed. (Run with `ZARR_V3_EXPERIMENTAL_API=1 pytest -vs`)

The next step is to turn this into a workflow that runs against Zarr's V3 branch (when it is created).

cc @MSanKeys963 @jhamman

tomwhite:
Zarr 3.0.0a0 was released last week. It's on [PyPI](https://pypi.org/project/zarr/3.0.0a0/), and can be installed by specifying `--pre` with `pip install`.

There's already a GH Actions workflow for this, which is currently failing (e.g. here's a [recent run](https://github.com/cubed-dev/cubed/actions/runs/9549809043/job/26320268341?pr=479)).

Some useful pre-requisites to getting Zarr V3 working are:
- [ ] #479
  - (Removes a dependency on an internal Zarr API)
- [ ] #441
  - (Removes a dependency on `BasicIndexer`, another internal Zarr API)
- [ ] #480
  - (Adds a mechanism to try out Zarr V3)

tomwhite:
Closing this as we have been running a CI job against Zarr v3 for a while now: https://github.com/cubed-dev/cubed/actions/workflows/zarr-v3-tests.yml.

PR #479, opened 2024-06-17
Add normalize_shape since it is not in Zarr's public API

tomwhite:


tomwhite:
Note that the **Tests / test (ubuntu-latest, 3.12, --pre) (pull_request)** failure is expected, see #295

PR #480, opened 2024-06-17
Pluggable backend storage

### LINKED ISSUE #322: Generalise storage layer
This would
1. allow the use of different storage backends in addition to the regular Zarr library - in particular #187.
2. convert the backend array API being used (NumPy, JAX, etc - see #317) to and from the underlying storage.

These belong together since they are quite closely related, e.g. the use case in #304 where the underlying storage is Zarr and the backend array is JAX on GPUs, so it makes sense to have something that encapsulates this.

tomwhite:
Fixes #322. This introduces a pluggable Zarr backend storage entrypoint (see 19b7dfeb00773fd3240d4b4b799ba1864b459c28), which will make it easier to adopt Zarr V3 (#295). Also enables #187.

### DIFF
--- a/cubed/core/ops.py
+++ b/cubed/core/ops.py
@@ -23,6 +23,7 @@ from cubed.primitive.blockwise import blockwise as primitive_blockwise
 from cubed.primitive.blockwise import general_blockwise as primitive_general_blockwise
 from cubed.primitive.rechunk import rechunk as primitive_rechunk
 from cubed.spec import spec_from_config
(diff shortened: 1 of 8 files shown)


tomwhite:
This is based on #479 and #441

tomwhite:
Rebased, and changed the configuration mechanism to be a top-level `storage_name` setting. We could add the ability to set the storage backend on a spec (so it can be changed for each computation) later if needed.

PR #484, opened 2024-06-19
Zarr v3

tomwhite:
Attempt to run against Zarr v3. There are still lots of failures - see #295


tomwhite:
This is passing against the upstream Zarr v3 branch, so I think it can be merged. There will be follow-on improvements once https://github.com/zarr-developers/zarr-python/issues/1978 is fixed.

PR #495, opened 2024-07-13
Fix: set allowed_mem above insufficient default

wietzesuijker:
This fixes the error I got: `ValueError: Projected blockwise memory (900000008) exceeds allowed_mem (200000000), including reserved_mem (100000000)`. The 8 gb is pretty arbitrary. At least the Spec addition makes it more explicit.


wietzesuijker:
The second commit feels like a band-aid fix, but the Spec didn't let me use the `max_mem` argument. It's not been fixed throughout the package though, just for a few examples.

tomwhite (commented):
Sorry that the examples didn't work smoothly for you @wietzesuijker, and thanks for submitting these improvements.

I think demo.ipynb was forgotten about which is why it's so out of date. As well as the `max_mem` change, I think we should change the Beam example (running locally) to be a Lithops one (again running locally) as that's what we tend to use most.

I'd be happy to fix that up, but happy for you to do that too if you want to?

tomwhite:
This should be fixed in #525. Please open an issue if you find any more problems @wietzesuijker. Thanks!

wietzesuijker:
Awesome, thanks @tomwhite !!

PR #503, opened 2024-07-18
Run tests on Python 3.13

tomwhite:


tomwhite:
Tests pass on Python 3.13 (with one exception, which seems to be a flaky test). This can be left until Python 3.13 is released.

PR #517, opened 2024-07-22
Check `allowed_mem` does not exceed total on machine for `processes` executor

### LINKED ISSUE #510: Check `allowed_mem` does not exceed total on machine for local executors
See #492

tomwhite:
Fixes #510


TomNicholas:
Yeah this is exactly the check I was thinking to add, nice!

Issue #277, opened 2023-07-24
Add scan / prefix sum primitive


dcherian:
If you're looking for something to do :), then scans would be a good thing to add.

Dask calls this "cumreduction" (terrible name!) : and its a quite useful primitive (xarray uses it for `ffill`, `bfill`). It's also a fun algorithm to think about: https://developer.nvidia.com/gpugems/gpugems3/part-vi-gpu-computing/chapter-39-parallel-prefix-sum-scan-cuda see the blelloch, 1990 section)

TomNicholas:
Oops I think I missed this one when creating the `ChunkManagerEntrypoint` in https://github.com/pydata/xarray/pull/701. Looks like calling xarray's `ffill` or `bfill` on an array with chunked core dimensions will [attempt to apply `dask.array.cumreduction`](https://github.com/pydata/xarray/blob/3fca02024a1a579c0bdd2bde66bfd0708a23c3e5/xarray/core/dask_array_ops.py#L87) along the core dims even if the underlying array is a `cubed.Array`.

tomwhite:
I think the relevant part of the [Nvidia doc](https://developer.nvidia.com/gpugems/gpugems3/part-vi-gpu-computing/chapter-39-parallel-prefix-sum-scan-cuda) is "39.2.4 Arrays of Arbitrary Size", which explains how to apply the algorithm to chunked (or blocked) arrays.

We could implement this by using the NumPy `cumsum` operation (or equivalent for the generalised operation) on the blocks, and then create the auxiliary arrays of sums (`SUMS` using the Nvidia terminology) and cumulative sums (`INCR`). 

Naively, `INCR` could have chunks of size one so that it could be mapped using a blockwise operation with the scanned blocks, but this would not scale well since if there were say 1000 chunks, then a single task would have to write 1000 tiny chunk files when storing `INCR` as a Zarr file.

Instead we could write `INCR` as a Zarr file with a single chunk, then a task processing a scanned block `i` would load the entire `INCR` file, but just use the `i`-th value to add to its chunk. This should work as long as the object store could handle many concurrent reads. Cubed has a `map_direct` operation that allows access to side inputs (`INCR`) for cases like this.

dcherian:
Ah yes that figure in "39.2.4" is what I remember. Such a cool algorithm!

Here's the dask PR where I learnt of this: https://github.com/dask/dask/pull/6675

dcherian:
I'm trying to do this to learn more about cubed. :)

[...]

PR #527, opened 2024-07-30
Tree merge chunks

tomwhite:
See #277


dcherian:
Shouldn't this be the algorithm in to `merge_chunks` for merging to an intermediate chunking?

tomwhite:
> Shouldn't this be the algorithm in to `merge_chunks` for merging to an intermediate chunking? I don't see why we'd want to restrict it to just condensing to one chunk along an axis.

You're right!

I'm going to close this for now as it's not needed for #531

PR #530, opened 2024-07-31
Update CUBED_CONFIG path in example README.md

thodson-usgs:
While testing cubed, I was experiencing a lot of new-user environment issues. One issue was that I'd setup `CUBED_CONFIG` to point to a directory rather than a `yaml` file, as is shown in README example. However, I was hacking on a problem and had several `yaml` files that I was testing. Depending on my naming convention, different files might get picked up as my `CUBED_CONFIG`. A better practice is to explicitly name the file in the README example.

Specifically, like

```shell
export CUBED_CONFIG=$(pwd)/processes/cubed.yaml
python add-asarray.py
```


tomwhite (approved):
LGTM. Thanks @thodson-usgs!

PR #508, opened 2024-07-20
Adding Jax tests for the M1 mac.

alxmrs:
I'm beginning to explore #304 in greater depth. Since the only local GPU I have access to is an M1 chip (I have an M1 Macbook Air), I thought I would replicate this environment in CI.

PR #536, opened 2024-08-03
Adding `compile_function` as execute option.

alxmrs:
This allows users to use jit or aot compilation during the dag finalization process within executors. It should work straightforwardly on jax/numba style jit compilation. It's possible, but maybe ugly, to perform jax-aot-style compilation.

This should help make progress towards #304.

(Note to reviewer: I realized this could be merged ahead of #508, so I pulled it out into a non-dependent PR).

### DIFF
--- a/cubed/tests/test_executor_features.py
+++ b/cubed/tests/test_executor_features.py
@@ -341,10 +341,25 @@ except ModuleNotFoundError:
 
 
 @pytest.mark.parametrize("compile_function", COMPILE_FUNCTIONS)
-def test_check_jit_compliation(spec, executor, compile_function):
+def test_check_compilation(spec, executor, compile_function):
     a = xp.asarray([[1, 2, 3], [4, 5, 6], [7, 8, 9]], chunks=(2, 2), spec=spec)
     b = xp.asarray([[1, 1, 1], [1, 1, 1], [1, 1, 1]], chunks=(2, 2), spec=spec)
     c = xp.add(a, b)
     assert_array_equal(
         c.compute(executor=executor, compile_function=compile_function), np.array([[2, 3, 4], [5, 6, 7], [8, 9, 10]])
     )
+
+
+def test_compilation_can_fail(spec, executor):
+    def compile_function(func):
+        raise NotImplementedError(f"Cannot compile {func}")
+
+    a = xp.asarray([[1, 2, 3], [4, 5, 6], [7, 8, 9]], chunks=(2, 2), spec=spec)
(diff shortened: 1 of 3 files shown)


alxmrs:
I think this PR is ready to merge. The next things to explore in this area are: 
- [ ] Testing AOT compilation
- [ ] Figuring out if this is the interface we need for exporting compiled functions (https://github.com/cubed-dev/cubed/issues/304#issuecomment-2258602859)

cubed/tests/test_executor_features.py
```diff
@@ -315,3 +315,66 @@ def test_check_runtime_memory_processes(spec, executor):
...
+try:
+    if 'jax' in os.environ.get('CUBED_BACKEND_ARRAY_API_MODULE', ''):
+        from jax import jit as jax_jit
+        COMPILE_FUNCTIONS.append(jax_jit)
+
+        def aot(func, *, config=None):
+            # TODO(alxmrs): implement lowering
+            return jax_jit(func)
+
+        COMPILE_FUNCTIONS.append(aot)
```
tomwhite, line 337: Maybe leave the AOT part out completely until it's working?

cubed/tests/test_executor_features.py
```diff
@@ -315,3 +315,66 @@ def test_check_runtime_memory_processes(spec, executor):
...
+
+    a = xp.asarray([[1, 2, 3], [4, 5, 6], [7, 8, 9]], chunks=(2, 2), spec=spec)
+    b = xp.asarray([[1, 1, 1], [1, 1, 1], [1, 1, 1]], chunks=(2, 2), spec=spec)
+    c = xp.add(a, b)
+    try:
+        c.compute(executor=executor, compile_function=compile_function)
+        assert False, "Compile function was not called."
+    except NotImplementedError as e:
+        assert True, "Compile function was applied."
+        assert "add" in str(e), "Compile function was applied to add operation."
```
tomwhite, line 365: [pytest.raises](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions) is more idiomatic for testing exceptions like this (and below).

tomwhite (commented):
This looks good - thanks for submitting it @alxmrs! Submitting the JAX work in smaller chunks is a good idea.

I made a couple of suggestions for changes, and it also looks like mypy is complaining.

(I'm interested in examples where compilation gives a performance increase - but that's not a blocker to merging this.)

tomwhite (approved):

tomwhite:
Thanks @alxmrs!

PR #545, opened 2024-08-08
Add `raise_if_computes` for testing

### LINKED ISSUE #310: Provide a way to check computations are lazy
Implement something like `raise_if_cubed_computes` similar to xarray's [`raise_if_dask_computes`](https://github.com/pydata/xarray/blob/8c21376aa94deb91cb10e78f56ecd68aa7fe90fa/xarray/tests/__init__.py#L125-L130)

tomwhite:
Fixes #310

This is useful for xarray testing.

I had hoped to be able to implement this by writing an executor that raises  when its `execute_dag` method is called - but that doesn't work. The problem is that the second part of this test fails, since `b`'s `spec` object contains the `raise-if-computes` executor since it was created within the context manager. Even though `compute` is called outside the context manager, the executor is still set on the `spec`, so that is what is used.

```python
def test_raise_if_computes():
    # shouldn't raise since compute has not been called
    with raise_if_computes():
        a = xp.ones((3, 3), chunks=(2, 2))
        b = xp.negative(a)

    # shouldn't raise since we are outside the context manager
    assert_array_equal(b.compute(), -np.ones((3, 3)))
```

I'm not sure that changing this is the right thing to do, hence this PR which does things a bit more directly.

### DIFF
--- a/cubed/core/array.py
+++ b/cubed/core/array.py
@@ -24,6 +24,10 @@ def gensym(name="array"):
     return f"{name}-{sym_counter:03}"
 
 
+# see cubed.testing
+compute_should_raise = False
+
+
 # type representing either a CoreArray or a public-facing Array
 T_ChunkedArray = TypeVar("T_ChunkedArray", bound="CoreArray")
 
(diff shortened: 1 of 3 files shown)


tomwhite:
I'm not too fond of this either, but it will probably need some more fundamental changes to how the spec/config is used to make the special executor/callback approach work. That would be a better solution though.

tomwhite:
Closing this for the time being

Issue #544, opened 2024-08-07
`TimelineVisualizationCallback()` usage [cubed - cubed/xarray]


norlandrhagen:
Related to this issue: https://github.com/cubed-dev/cubed/issues/543. I'm trying out some of the callbacks in timeline. Once again, might be getting the usage totally wrong. 

```python

from cubed import Spec
import xarray as xr 
from cubed.diagnostics.timeline import TimelineVisualizationCallback

spec = Spec(work_dir='tmp', allowed_mem='2GB')
timeline_viz = TimelineVisualizationCallback()

ds = xr.tutorial.open_dataset('air_temperature', chunked_array_type='cubed',
     from_array_kwargs={'spec': spec},chunks={})


ds.compute(callbacks=[timeline_viz])

```

When I call this, I get a `FileNotFoundError`:

```
FileNotFoundError: [Errno 2] No such file or directory: '~/cubed/notebooks/history/compute-20240807T124600.864122/timeline.svg'
```
Full traceback:

<details>

```python

FileNotFoundError                         Traceback (most recent call last)
Cell In[23], [line 12](vscode-notebook-cell:?execution_count=23&line=12)
      [6](vscode-notebook-cell:?execution_count=23&line=6) timeline_viz = TimelineVisualizationCallback()
      [8](vscode-notebook-cell:?execution_count=23&line=8) ds = xr.tutorial.open_dataset('air_temperature', chunked_array_type='cubed',
      [9](vscode-notebook-cell:?execution_count=23&line=9)      from_array_kwargs={'spec': spec},chunks={})
---> [12](vscode-notebook-cell:?execution_count=23&line=12) ds.compute(callbacks=[timeline_viz])

File ~/micromamba/envs/cubed/lib/python3.11/site-packages/xarray/core/dataset.py:1038, in Dataset.compute(self, **kwargs)
   [1014](https://file+.vscode-resource.vscode-cdn.net/Users/nrhagen/Documents/carbonplan/cubed/notebooks/~/micromamba/envs/cubed/lib/python3.11/site-packages/xarray/core/dataset.py:1014) """Manually trigger loading and/or computation of this dataset's data
   [1015](https://file+.vscode-resource.vscode-cdn.net/Users/nrhagen/Documents/carbonplan/cubed/notebooks/~/micromamba/envs/cubed/lib/python3.11/site-packages/xarray/core/dataset.py:1015) from disk or a remote source into memory and return a new dataset.
[...]

Issue #543, opened 2024-08-07
`HistoryCallback()` in .compute [cubed / cubed-xarray]


norlandrhagen:
Thanks for the new diagnostics docs @tomwhite! 

Trying to get my head around using `HistoricalCallback()` in .compute(). One example that uses it seems to be in [examples/add-random.py](https://github.com/cubed-dev/cubed/blob/bdf6dba80ecbb40efcd28c2132bce4a49fcbdc73/examples/add-random.py#L20).  Should I expect to see any output when adding `HistoryCallback()` to the callbacks in `.compute` or is this the wrong usecase?

```python 
from cubed import Spec
import xarray as xr 
from cubed.diagnostics.history import HistoryCallback

spec = Spec(work_dir='tmp', allowed_mem='2GB')
hist = HistoryCallback()

ds = xr.tutorial.open_dataset('air_temperature', chunked_array_type='cubed',
     from_array_kwargs={'spec': spec},chunks={})

rds = ds.chunk({'time':1}, chunked_array_type="cubed")

rds.compute(callbacks=[hist])
```

TomNicholas:
Doesn't `HistoryCallback` save a bunch of extra csv files into the `tmp` directory? (Or maybe just the local directory? Pretty sure that's what it did back when I wrote the blog post https://xarray.dev/blog/cubed-xarray

tomwhite:
It will save some CSV files in a `history/compute-{id}` directory. For example:

```shell
> ls history/compute-20240807T141305.118660                      
events.csv   plan.csv     stats.csv    timeline.svg
```

The `timeline.svg` file is created by the `TimelineVisualizationCallback`. These files are referred to in https://github.com/cubed-dev/cubed/tree/main/examples#examples-1, but I think the documentation needs to be expanded.

norlandrhagen:
Oh good to know! I can create a small PR to add this to the docs if that's helpful.

tomwhite:
> Oh good to know! I can create a small PR to add this to the docs if that's helpful.

That would be great, thanks @norlandrhagen!

norlandrhagen:
https://github.com/cubed-dev/cubed/issues/538

PR #548, opened 2024-08-08
Adds logging [INFO] to `TimelineVisualizationCallback` and `HistoryCallback` 

norlandrhagen:
Small PR that:

- Adds info level logging to show the directory that `TimelineVisualizationCallback` and `HistoryCallback` write their artifacts to.
- Updates the `TimelineVisualizationCallback` os.mkdir to the Pathlib mkdir used in the HistoryCallback. This seems to fix the `FileNotFoundError` in https://github.com/cubed-dev/cubed/issues/544.

example logging:
```
TimelineVisualizationCallback results saved to directory: history/compute-20240808T135430.657101/
HistoryCallback results saved to directory: history/compute-20240808T135430.657101/
```
Some remaining Q's:
- The `TimelineVisualizationCallback` calls an external function (to the class) `create_timeline` that has args for such as `dst=None`, which are not a part of `TimelineVisualizationCallback`. Is this `create_timeline` function used anywhere else? I ask this b/c the optional `dst` arg creates a few more path creation options. If it is confined to `TimelineVisualizationCallback`, the logic could be more similar to `HistoryCallback` which is a bit cleaner. 

- I added info level logging to `diagnostics/__init__.py` is this OK to set logging here?

### DIFF
--- a/docs/user-guide/diagnostics.md
+++ b/docs/user-guide/diagnostics.md
@@ -41,7 +41,12 @@ Array `c` is coloured orange, which means it is materialized as a Zarr array. Ar
 
 Similarly, the operation that produces `c` is shown in a lilac colour to signify that it runs tasks to produce the output. Operations `op-001` and `op-002` don't run any tasks since `a` and `b` are just small constant arrays.
 
-## Progress bar
+
+## Callbacks
+
+You can pass callbacks to functions that call `compute`, such as {py:func}`store <cubed.store>` or {py:func}`to_zarr <cubed.to_zarr>`.
+
+### Progress bar
 
 You can display a progress bar to track your computation by passing callbacks to {py:meth}`compute() <cubed.Array.compute()>`:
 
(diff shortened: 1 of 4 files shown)


TomNicholas:
Nice! Would this close either of #543 or #544?

norlandrhagen:
I think it should close both!

norlandrhagen:
No idea why the Apache Beam tests are failing though 😕

tomwhite:
Thanks for the PR @norlandrhagen!

> * The `TimelineVisualizationCallback` calls an external function (to the class) `create_timeline` that has args for such as `dst=None`, which are not a part of `TimelineVisualizationCallback`. Is this `create_timeline` function used anywhere else? I ask this b/c the optional `dst` arg creates a few more path creation options. If it is confined to `TimelineVisualizationCallback`, the logic could be more similar to `HistoryCallback` which is a bit cleaner.

This code was originally based on code from Lithops for producing a timeline image, so some of it may be redundant. Please go ahead and remove code paths that are not used so it's more similar to `HistoryCallback`.

> * I added info level logging to `diagnostics/__init__.py` is this OK to set logging here?

No, it should go in client code. Perhaps add it to examples/add-random.py and examples/matmul-random.py?

tomwhite:
> No idea why the Apache Beam tests are failing though 😕

Not related to this PR. Opened #549

norlandrhagen:
> > * I added info level logging to `diagnostics/__init__.py` is this OK to set logging here?
> 
> No, it should go in client code. Perhaps add it to examples/add-random.py and examples/matmul-random.py?

Do you see any place where logging or optional logging might be useful? Maybe default off, but user enabled --debug flag could output some info about where callback artifacts are written. Also happy to remove it and just mention in the docs of where you can expect them to be written. 

ex:
```
TimelineVisualizationCallback results saved to directory: history/compute-20240809T115943.795332/timeline.svg/
HistoryCallback results saved to directory: history/compute-20240809T115943.795332/
```

tomwhite:
It would be useful, but the problem is that if you configure logging then that will interfere with logging that the user has configured (often in non-obvious ways). So I think adding logging to the examples is a good compromise. Also mentioning in the docs.

norlandrhagen:
Makes sense! Thanks for explaining.

docs/user-guide/diagnostics.md
```diff
@@ -53,14 +58,37 @@ You can display a progress bar to track your computation by passing callbacks to
...
+
 This will work in Jupyter notebooks, and for all executors.
 
-You can also pass callbacks to functions that call `compute`, such as {py:func}`store <cubed.store>` or {py:func}`to_zarr <cubed.to_zarr>`.
 
-## History and timeline visualization
 
-The history and timeline visualization callbacks can be used to find out how long tasks took to run, and how much memory they used.
+### History
+The history callback can be used to understand how long tasks took to run, and how much memory they used. The history callback will write [`events.csv`, `plan.csv` and `stats.csv`] to a new directory under with the schema `history/compute-{id}`.
```
tomwhite, line 71:
```suggestion
The history callback can be used to understand how long tasks took to run, and how much memory they used. The history callback will write [`events.csv`, `plan.csv` and `stats.csv`] to a new directory under the current directory with the schema `history/compute-{id}`.
```

tomwhite (approved):
Thanks @norlandrhagen! LGTM. I've made one small suggestion. Let me know when you're ready to merge.

norlandrhagen:
If you're happy with the doc suggestions, then it's ready to merge.

PR #551, opened 2024-08-13
Cast inputs to Cubed arrays in `apply_ufunc`

tomwhite:


cubed/core/gufunc.py
```diff
@@ -68,7 +68,11 @@ def apply_gufunc(
     # Main code:
 
     # Cast all input arrays to cubed
-    # args = [asarray(a) for a in args]  # TODO: do we need to support casting?
+    from cubed.array_api.creation_functions import asarray
+
+    specs = [a.spec for a in args if hasattr(a, "spec")]
+    spec = specs[0] if len(specs) > 0 else None
```
TomNicholas, line 74: Should multiple different specs error? Or specs with different values for the allowed memory and intermediate directory. Otherwise isn't it possible to get some confusing behaviour where it isn't super clear which memory configuration and output directory is being used?

tomwhite: This actually gets checked later when constructing the plan (all specs have to be the same), but I should add a comment to explain that.

PR #555, opened 2024-08-18
Example using "dragon" start method

applio:
In discussions with Tom and Tom, it was unclear where this example involving the use of the `"dragon"` start method should land.  This PR is intended to help further those discussions and ensure we do not lose track of this working example.

This PR is dependent upon PR #554 and is not intended to be merged independent of that PR.

### DIFF
--- a/cubed/runtime/executors/local.py
+++ b/cubed/runtime/executors/local.py
@@ -180,7 +180,10 @@ async def async_execute_dag(
         check_runtime_memory(spec, max_workers)
     if use_processes:
         max_tasks_per_child = kwargs.pop("max_tasks_per_child", None)
-        context = multiprocessing.get_context("spawn")
+        if isinstance(use_processes, str):
+            context = multiprocessing.get_context(use_processes)
+        else:
+            context = multiprocessing.get_context("spawn")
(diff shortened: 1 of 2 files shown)


tomwhite:
I'm still not sure where the best place to add an example would be. The Dragon documentation would be one option (with a link from the Cubed documentation and examples). Or it could go here, but with the script named `matmul-random-dragon.py` or something so that it is standalone. We'd also want a link to how to set up Dragon.

@TomNicholas you've been trying out Dragon so perhaps you have thoughts?

tomwhite:
Thinking about this more, it should be possible to add a `dragon/cubed.yaml` file under [`examples`](https://github.com/cubed-dev/cubed/tree/main/examples) and then run

```shell
export CUBED_CONFIG=$(pwd)/dragon/cubed.yaml
python add-asarray.py
```

This would make it just like the other executors and allow any of the examples to be run on Dragon.

TomNicholas:
@applio we talked about this at SciPy this year - are you or Colin (whose github handle I don't know) planning to finish this off?

The other thing we discussed that we should probably do is add a test using Dragon to the Cubed CI. That should be simple if we can just pip install Dragon and run it on a single machine.

PR #558, opened 2024-08-21
Memray integration

### LINKED ISSUE #372: Profile Cubed task memory usage with Memray
I've been using [Memray](https://bloomberg.github.io/memray/) to profile the memory usage of Cubed tasks running locally using a local Lithops executor (since it runs tasks in a separate process).

This give great insight into how memory is allocated for blockwise operations. Here is an example timeline for an array `add` operation, where you can see Zarr chunks being loaded and stored:

memray-add

More details in [this presentation](https://docs.google.com/presentation/d/1YWT8PAAMEK65rMlT3XFwwxjZE0vvl5KHhwpZjHLUxtk/edit#slide=id.p).

This issue is to add the instrumentation code to make it easy to profile any computation.

tomwhite:
Fixes #372


docs/user-guide/diagnostics.md
```diff
@@ -92,3 +92,41 @@ The timeline callback will write a graphic `timeline.svg` to a directory with th
...
+
+```shell
+(cd $(ls -d history/compute-* | tail -1)/memray; for f in $(ls *.bin); do echo $f; python -m memray flamegraph --temporal -f -o $f.html $f; done)
+```
+
+Here is the flame graph for the add operation:
+
+![Memray temporal view of an 'add' operation](../images/memray-add.png)
+
+Annotations have been added to exaplin what is going on in this example. Note that reading a chunk from Zarr requires twice the chunk memory (400MB) since there is a buffer for the compressed Zarr block (200MB), as well as the resulting array (200MB). After the first chunk has been loaded the memory dips back to 200MB since the compressed buffer is no longer retained.
```
TomNicholas, line 132: spelling
```suggestion
Annotations have been added to explain what is going on in this example. Note that reading a chunk from Zarr requires twice the chunk memory (400MB) since there is a buffer for the compressed Zarr block (200MB), as well as the resulting array (200MB). After the first chunk has been loaded the memory dips back to 200MB since the compressed buffer is no longer retained.
```

TomNicholas:
Cool! 

This could pair very well with an `apply_gufunc` example (https://github.com/cubed-dev/cubed/issues/519). The docs could guide you through how to apply a custom ufunc (e.g. something from `scipy`), and measure its memory usage locally, so that you know how big your containers in the cloud would need to be.

Issue #565, opened 2024-09-02
Rewrite "why cubed?" for users


TomNicholas:
The current text on [why cubed?](https://cubed-dev.github.io/cubed/getting-started/why-cubed.html) is really written at a very high level. It's very interesting for people who think about different distributed systems, but it's not super clear to new users who want to know what cubed can do for them.

Perhaps we should move the existing text to e.g. `design/motivation`, and rewrite this intro text to instead emphasise:
- cubed is a drop-in replacement for `dask.array`
- it tells you if your computation would run out of memory
- it is designed to be very robust to failures, and be resumable
- it doesn't require you to make many (if any) up-front choices about configuration / cluster size
- (Optional) IF you want to run in the cloud it is very horizontally scalable and stateless / serverless

tomwhite:
Thanks for opening this @TomNicholas!

> Perhaps we should move the existing text to e.g. `design/motivation`

Definitely. It's mainly of historical interest, so should be moved down to the bottom of the nav links (I noticed that [DuckDB does this](https://duckdb.org/docs/)).

The intro text at https://cubed-dev.github.io/cubed/ (and https://github.com/cubed-dev/cubed/blob/main/README.md) needs updating, as you say.
* "[Zarr](https://zarr.readthedocs.io/en/stable/) for storage" is a bit vague. It's about how Zarr is used for intermediate storage, which is an implementation detail that is probably not a top-level bullet point. Being able to read and write the user's Zarr data is more relevant.
* "Multiple serverless runtimes: Python (in-process)" - change to add out-of-core/processes.
  * Maybe say something like "Multiple runtimes: local (...), serverless (...), cluster (Beam, Dask)"
* Add something about accelerators (JAX)

These are just a few ideas. I think it would be really useful to collaborate on re-writing the intro.

tomwhite:
Fixed in #622

PR #622, opened 2024-11-18
Rewrite Cubed intro docs

tomwhite:
Here's my first attempt at #565. Appreciate any thoughts you may have @TomNicholas and @rabernat.

Screenshot 2024-11-18 at 12 58 51

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,20 +1,14 @@
 # Cubed
 
-## Bounded-memory serverless distributed N-dimensional array processing
+## Scalable out-of-core array processing in Python
 
-Cubed is a distributed N-dimensional array library implemented in Python using bounded-memory serverless processing and Zarr for storage.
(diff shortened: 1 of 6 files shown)


TomNicholas:
On the tagline:

I think you need to say "bounded-memory" as that is really the USP, and I think you need at least one other word to convey the fact cubed can scale. That could potentially be "scalable", "out-of-core", or "distributed". I don't have a strong opinion beyond that, but I think you need both of those to communicate what cubed offers.

On the rest of the layout: We discussed this in the meet last week and I think it looks great! I think the 6 main boxes are well-chosen features to highlight.

tomwhite:
Thanks Tom - excellent points.

I'm thinking of updating the tagline to "Scalable array processing with bounded memory".

PR #623, opened 2024-11-20
Support `peak_measured_mem` on Windows

### LINKED ISSUE #177: Support `peak_measured_mem` on Windows
With #176 you can run Cubed on Window, but `peak_measured_mem` isn't implemented.

To implement we could use `psutil`, but we should make it optional so that other platforms aren't required to import it. 

See https://pythonspeed.com/articles/estimating-memory-usage/#measuring-peak-memory-usage

tomwhite:
Fixes #177 

This is needed for #621

### DIFF
--- a/cubed/tests/test_core.py
+++ b/cubed/tests/test_core.py
@@ -1,4 +1,3 @@
-import platform
 import random
 from functools import partial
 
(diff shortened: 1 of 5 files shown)


tomwhite:
This is failing for Windows on Lithops, since Lithops doesn't support measuring memory on Windows (see https://github.com/lithops-cloud/lithops/blob/fdb6432e9fdc1191a4205579efc0bac32bd8daf7/lithops/worker/utils.py#L151-L152). In addition, it seems to be returning "None" (as a string), which looks like a Lithops bug that should be reported upstream.

PR #624, opened 2024-11-20
Remove direct dependency on fsspec

tomwhite:
It is not used directly in Cubed, and Zarr has it as a transitive dependency.


tomwhite:
> Zarr has it as a transitive dependency

On further investigation this is only true for zarr-python v3. However, Cubed does not use fsspec directly, except in its tests. So it would need more work to remove it. I'm going to close this for the moment.

PR #633, opened 2024-12-04
Icechunk store

### LINKED ISSUE #628: Icechunk integration
It would be great to read and write data from [Icechunk](https://icechunk.io/).

Reads should just use [`cubed.from_zarr`](https://cubed-dev.github.io/cubed/generated/cubed.from_zarr.html). Writes are more difficult, as you have to handle changesets (see https://github.com/earth-mover/icechunk/pull/357).

tomwhite:
Fixes #628

This is a proof of concept/draft for a `store_icechunk` function that uses Cubed callbacks to do the Icechunk store merging. The tests pass locally, but not in CI yet. I'm not sure where this code should live.

### DIFF
--- a/cubed/icechunk.py
+++ b/cubed/icechunk.py
@@ -32,10 +32,6 @@ def store_icechunk(
             f"Different number of sources ({len(sources)}) and targets ({len(targets)})"
         )
 
-    if isinstance(sources, CoreArray):
-        sources = [sources]
-        targets = [targets]
-
     arrays = []
     for source, target in zip(sources, targets):
         identity = lambda a: a
--- a/cubed/icechunk.py
+++ b/cubed/icechunk.py
@@ -1,7 +1,7 @@
 from typing import TYPE_CHECKING, Any, List, Sequence, Union
 
 import zarr
-from icechunk import IcechunkStore
+from icechunk import Session
 
 from cubed import compute
 from cubed.core.array import CoreArray
@@ -13,7 +13,7 @@ if TYPE_CHECKING:
 
 
 def store_icechunk(
-    store: IcechunkStore,
+    session: Session,
     *,
     sources: Union["Array", Sequence["Array"]],
     targets: List[zarr.Array],
@@ -48,7 +48,7 @@ def store_icechunk(
         )
         arrays.append(array)
 
-    # use a callback to merge icechunk stores
+    # use a callback to merge icechunk sessions
     store_callback = IcechunkStoreCallback()
     # add to other callbacks the user may have set
     callbacks = kwargs.pop("callbacks", [])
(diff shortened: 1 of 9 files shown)


cubed/icechunk.py
```diff
@@ -0,0 +1,86 @@
...
+if TYPE_CHECKING:
+    from cubed.array_api.array_object import Array
+
+
+def store_icechunk(
+    store: IcechunkStore,
+    *,
+    sources: Union["Array", Sequence["Array"]],
+    targets: List[zarr.Array],
+    executor=None,
```
dcherian, line 20: could definitely address this later but `regions` is quite an important kwarg.

cubed/icechunk.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def on_task_end(self, event):
+        result = event.result
+        if result is None:
+            return
+        for store in result:
+            if self.store is None:
+                self.store = store
+            else:
+                self.store.merge(store.change_set_bytes())
```
dcherian, line 86: @paraseba shall we commit to `icechunk.distributed.merge_stores` as public API?

https://github.com/earth-mover/icechunk/blob/a8c33b81d329c32a8e34e7151a1a71620967067a/icechunk-python/python/icechunk/distributed.py#L12-L16

dcherian: @tomwhite It looks like callbacks are "accumulated" on every worker. Is that right?

dcherian: Right we can fix that, but note that dask and presumably cubed are accumulating on remote workers already,  so there is already some parallelism in how it is used.

Moreover it'd be nice not to have `store.change_set_bytes` be the public API :)

dcherian: > but would it be possible to merge a batch in one go rather than one at a time

we'll have to build some rust API, we'll get to this eventually.

> No, they are being accumulated on the client so there is no parallelism. 

ah ok. In that case, how about calling `reduction` on each array that was written? That way you parallelize the merge across blocks for each array, and then the only serial bit is the merging across arrays, which will be a lot smaller. I considered this approach for dask, but then just wrote out a tree reduction across all chunks.

dcherian: > Yes, that's basically it

:( I was afraid so.

tomwhite:
> I'm not sure where this code should live.

After chatting to @rabernat, @jhamman, @TomNicholas we think Icechunk integration can live in this repo - as an optional dependency. There may be further changes once Xarray supports pluggable writers, but this PR can go in once tests are passing.

rabernat:
This is amazing!

Note that the Python API is getting a major refactor in https://github.com/earth-mover/icechunk/pull/470 which will almost certainly break this.

tomwhite:
> Note that the Python API is getting a major refactor in [earth-mover/icechunk#470](https://github.com/earth-mover/icechunk/pull/470) which will almost certainly break this.

@rabernat thanks for the heads up. I'll keep an eye on that PR.

I think the best place to run the Icechunk tests here is in the Zarr v3 tests workflow in CI.

tomwhite:
This is working well now with the latest zarr and icechunk releases. I plan to merge it soon.

PR #636, opened 2024-12-04
Array API tests fixes

tomwhite:
Fix some failures in https://github.com/cubed-dev/cubed/actions/workflows/array-api-tests.yml (there may be more).


.github/workflows/array-api-tests.yml
```diff
@@ -108,6 +108,8 @@ jobs:
           # edge case failures (https://github.com/cubed-dev/cubed/issues/420)
           array_api_tests/test_linalg.py::test_tensordot
           array_api_tests/test_linalg.py::test_vecdot
+          # (getitem with negative step size is not implemented)
```
dcherian, line 111: :) we actually rewrite such queries in Xarray for Zarr

tomwhite: Oh interesting! How do you do that?

dcherian: https://github.com/pydata/xarray/blob/99ee8c6ca54057a9b994d7685f36236f2d5a69d9/xarray/core/indexing.py#L1084 and friends. We rewrite the query to normal slice with +ve stride, then reverse in-memory after read :)

This is one of Xarray's magic tricks that very few people know about. We guarantee consistent indexing API over any array for the most part.

In https://github.com/pydata/xarray/issues/8667, I suggested reusing this machinery for cubed

tomwhite: Very cool. Would be great to reuse that. BTW Cubed `flip` is implemented in a similar way, but it's only `step=-1` of course.

tomwhite: Opened #637 to track

PR #520, opened 2024-07-25
WIP: Add virtual-rechunk example

thodson-usgs:
# Rechunk a virtual dataset

This example demonstrates how to rechunk a collection of necdf files on s3 into a single zarr store.

First, lithops and VirtualiZarr construct a virtual dataset comprised of the netcdf files on s3. Then, xarray-cubed rechunks the virtual dataset into a zarr. Inspired by the [Pythia cookbook](https://projectpythia.org/kerchunk-cookbook) by @norlandrhagen.

## STATUS
Fails during the chunk operation. Hoping someone with more experience might catch an obvious error on my part.

```bash
2024-07-25 09:19:10,806 [WARNING] future.py:254 -- ExecutorID 7a9116-3 | JobID M002 - CallID: 00000 - There was an exception - Activation ID: 65c097fe-6362-4aeb-8a77-074977f98559 - FileNotFoundError
2024-07-25 09:19:10,806 [INFO] invokers.py:107 -- ExecutorID 7a9116-3 | JobID M006 - Selected Runtime: virtualizarr-runtime - 2000MB
2024-07-25 09:19:10,806 [INFO] invokers.py:174 -- ExecutorID 7a9116-3 | JobID M006 - Starting function invocation: <lambda>() - Total: 1 activations
2024-07-25 09:19:10,806 [INFO] invokers.py:213 -- ExecutorID 7a9116-3 | JobID M006 - View execution logs at /private/var/folders/zr/jc46d6cs531df81bpgj4681sz25hhs/T/lithops-thodson/logs/7a9116-3-M006.log
```
The execution log complains  `FileNotFoundError: [Errno 2] No such file or directory: '/function/combined.json/.zmetadata'`:
```bash
Activation: 'virtualizarr-runtime' (445a08aa-60ba-48db-a3ee-659be326444f)
[
    2024-07-25 14:19:08,140 [INFO] handler.py:177 -- Lithops v3.4.1 - Starting AWS Lambda execution
    2024-07-25 14:19:08,140 [INFO] handler.py:178 -- Execution ID: 7a9116-3-M005/00000
    2024-07-25 14:19:08,243 [INFO] aws_s3.py:59 -- S3 client created - Region: us-west-2
    ----------------------- EXCEPTION !-----------------------
    Traceback (most recent call last):
      File "/function/lithops/worker/jobrunner.py", line 210, in run
        func = pickle.loads(self.job.func)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #646, opened 2024-12-17
VirtualiZarr example

tomwhite:
This is the example I was working on at the [Post-AGU Pangeo Hack Day](https://discourse.pangeo.io/t/post-agu-pangeo-hack-day-working-meeting-december-14-2024-in-washington-dc/4440) on Saturday.

The idea is to take multiple NetCDF files, combine with VirtualiZarr, and then rechunk using Cubed. This works, although there are some rough edges still. Also, when I tried using Icechunk rather than Kerchunk to write the references, I got an error when reading them back, so that needs investigating too.


tomwhite:
This has a lot of overlap with #520

tomwhite:
It would be good to restart this and get it published. It currently write references to Kerchunk format, but is this what you would recommend @TomNicholas, or is Icechunk a better default choice now?

PR #660, opened 2025-01-10
Pin Zarr v2 by default

### LINKED ISSUE #657: Test on Zarr Python v2 and v3
Now that Zarr Python 3.0.0 is out we are no longer testing on v2. We should have CI tests for both v2 and v3.

tomwhite:
Fixes #657

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,7 +2,7 @@
 omit =
     */tests/*
     cubed/array_api/*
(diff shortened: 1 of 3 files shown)


tomwhite:
This will help us avoid getting blocked with Zarr Python v3 changes like #662. At a future date we can remove this so that Zarr Python v3 is the default.

PR #666, opened 2025-01-13
Using `default_dtypes` instead of hard-coding dtypes.

### LINKED ISSUE #658: Use `default_dtypes` to determine the output data type in array API implementations
Currently we hard-code these. For example:

https://github.com/cubed-dev/cubed/blob/58ea08b8b578a486efaad6844947f30bb19db18c/cubed/array_api/creation_functions.py#L105-L111

Now we have [`default_dtypes`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.info.default_dtypes.html#array_api.info.default_dtypes) in [Cubed](https://github.com/cubed-dev/cubed/blob/58ea08b8b578a486efaad6844947f30bb19db18c/cubed/array_api/inspection.py#L17-L18), we can use it to get the default dtype in cases like this.

alxmrs:
Fixes #658.

### DIFF
--- a/cubed/array_api/creation_functions.py
+++ b/cubed/array_api/creation_functions.py
@@ -12,7 +12,7 @@ from cubed.storage.virtual import (
 )
 from cubed.utils import normalize_shape, to_chunksize
 from cubed.vendor.dask.array.core import normalize_chunks
-from cubed.array_api import __array_namespace_info__, closest_default_dtype
+from cubed.array_api import __array_namespace_info__
 
 if TYPE_CHECKING:
     from .array_object import Array
--- a/cubed/array_api/creation_functions.py
+++ b/cubed/array_api/creation_functions.py
@@ -58,9 +58,8 @@ def arange(
     num = int(max(math.ceil((stop - start) / step), 0))
     if dtype is None:
         dtype = nxp.arange(start, stop, step * num if num else step).dtype
-        # # the default nxp call does not adjust the data type to the default precision.
-        # dtype = __array_namespace_info__.dtypes(dtype.kind, device=device)
-        # dtype = to_default_precision(dtype, device=device)
+        # the default nxp call does not adjust the data type to the default precision.
+        dtype = closest_default_dtype(dtype, device=device)
 
     chunks = normalize_chunks(chunks, shape=(num,), dtype=dtype)
     chunksize = chunks[0][0]
--- a/cubed/array_api/creation_functions.py
+++ b/cubed/array_api/creation_functions.py
@@ -12,26 +12,42 @@ from cubed.storage.virtual import (
 )
 from cubed.utils import normalize_shape, to_chunksize
 from cubed.vendor.dask.array.core import normalize_chunks
-from cubed.array_api import __array_namespace_info__
+from cubed.array_api import __array_namespace_info__, closest_default_dtype
 
 if TYPE_CHECKING:
     from .array_object import Array
 
 
+def _traverse(it):
+    """Iterate over arbitrarily nested collections."""
+    stack = [it]
+
+    while stack:
+        item = stack.pop()
+        if isinstance(item, Iterable):
+            stack.extend(reversed(list(item)))  # Extend with reversed order to maintain original order
+        else:
+            yield item
+
+
 def _iterable_to_default_dtype(it, device=None):
     """Determines the default precision dtype of a collection (of collections) of scalars"""
-    w = it
-    while isinstance(w, Iterable):
-        w = next(iter(w))
     dtypes = __array_namespace_info__().default_dtypes(device=device)
-    if nxp.issubdtype(type(w), nxp.integer):
+
+    # Collect all the data types in the (arbitrarily) nested collection,
+    # and choose the highest promoted data type.
+    all_dtypes = {type(x) for x in _traverse(it)}
+    highest_type = nxp.result_type(*all_dtypes)
+
+    # TODO(alxmrs): Is there a better way to do this?
+    if nxp.issubdtype(highest_type, nxp.integer):
         return dtypes["integral"]
-    elif nxp.isreal(w):
-        return dtypes["real floating"]
-    elif nxp.iscomplex(w):
+    elif nxp.issubdtype(highest_type, nxp.complexfloating):
         return dtypes["complex floating"]
+    elif nxp.issubdtype(highest_type, nxp.floating):
+        return dtypes["real floating"]
     else:
-        raise ValueError(f"there are no default data types supported for {it}.")
+        return highest_type
 
 
 def arange(
@@ -83,11 +99,13 @@ def asarray(
     ):  # pragma: no cover
         return asarray(a.data)
     elif not isinstance(getattr(a, "shape", None), Iterable):
-        dtype = _iterable_to_default_dtype(a, device=device)
+        if dtype is None:
+            dtype = _iterable_to_default_dtype(a, device=device)
         a = nxp.asarray(a, dtype=dtype)
 
     if dtype is None:
-        dtype = a.dtype
+        dtype = closest_default_dtype(a.dtype, device=device)
+        a = a.astype(dtype)
 
     chunksize = to_chunksize(normalize_chunks(chunks, shape=a.shape, dtype=dtype))
     name = gensym()
--- a/cubed/array_api/creation_functions.py
+++ b/cubed/array_api/creation_functions.py
@@ -1,6 +1,7 @@
 import math
 from typing import TYPE_CHECKING, Iterable, List
 
+from cubed.array_api import __array_namespace_info__
 from cubed.backend_array_api import namespace as nxp
 from cubed.core import Plan, gensym
 from cubed.core.ops import map_blocks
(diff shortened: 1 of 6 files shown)


cubed/array_api/creation_functions.py
```diff
@@ -62,10 +98,13 @@ def asarray(
     ):  # pragma: no cover
         return asarray(a.data)
     elif not isinstance(getattr(a, "shape", None), Iterable):
-        # ensure blocks are arrays
+        if dtype is None:
+            dtype = _iterable_to_default_dtype(a, device=device)
```
tomwhite, line 102: Why do we need to find the dtype here at all - rather than let `nxp.asarray` in the following line resolve it?

cubed/array_api/inspection.py
```diff
@@ -22,3 +22,11 @@ def devices(self):
 
     def dtypes(self, *, device=None, kind=None):
         return nxp.__array_namespace_info__().dtypes(device=device, kind=kind)
+
+
+def closest_default_dtype(dtype, *, device=None):
```
tomwhite, line 27: This is not a part of the Array API, so should probably live somewhere else? Also it could do with a unit test.

tomwhite (commented):
Thanks for working on this @alxmrs! Added a few comments.

alxmrs:
Thanks for the early feedback, Tom. I'll TAL later today.

cubed/array_api/inspection.py
alxmrs: Thanks. I see now how everything in the package is part of the Array API. I'm moving this as a private utility function in the create module for now. If it's needed in more places, I'll find a better place for it later.

tomwhite:
This is looking good!

BTW I just noticed that we don't enforce linting in CI (I opened #668), so you might want to install pre-commit in your dev env and run `pre-commit run --all-files`.

alxmrs:
Thanks for pointing out how to lint. 

As a next step for this PR, I was going to add `default_dtype` logic to other places besides creation functions. Does that make sense? Would you recommend against this?

tomwhite:
> As a next step for this PR, I was going to add `default_dtype` logic to other places besides creation functions.

That would be great!

alxmrs:
I just found this useful dictionary! It's really good to know that these are defined somewhere we can access, I've been looking for just the thing w/in the array api!

https://github.com/cubed-dev/cubed/blob/5f75ba24544fb3f16ab7b8f477334fa47da1a1b2/cubed/array_api/dtypes.py#L78

alxmrs:
Hey @tomwhite, I have a few questions for what types to use in my latest commit. For example, see here: 
https://github.com/cubed-dev/cubed/pull/666/commits/8ccb3439e63043288b18bfcd17edf29e3aa58e61#diff-36cd7306bc9a90e7b718d59616eaf4c80aa06e61e9f1d6262eecc37ad36a4ceeR37

On one hand, it makes sense to me that these should be `default_dtype`s, since certain types won't work on specific hardware. On the other hand, I don't want to truncate intermediary values. 

In my next commit, I'm going to make these default dtypes. If that's not the way we should go, I'll just undo the commit.

alxmrs:
(looks like the next next step is to add more test coverage 😬 .)

tomwhite:
> On one hand, it makes sense to me that these should be `default_dtype`s, since certain types won't work on specific hardware. On the other hand, I don't want to truncate intermediary values.
> 
> In my next commit, I'm going to make these default dtypes. If that's not the way we should go, I'll just undo the commit.

That's a very good point. The choice of dtype for intermediate values is an implementation choice - and is not tied down by the spec. I'm inclined to not change this now in this PR, and look at it when looking at a concrete case for (say) JAX on GPUs. It would be good to get this merged as a relatively straightforward change - even if some cases need follow-ups.

cubed/array_api/creation_functions.py
```diff
@@ -210,15 +220,23 @@ def linspace(
...
+):
+    dtypes = __array_namespace_info__().default_dtypes(device=device)
+
     bs = x.shape[0]
     i = block_id[0]
     adjusted_bs = bs - 1 if endpoint else bs
-    blockstart = start + (i * size * step)
-    blockstop = blockstart + (adjusted_bs * step)
+    # While the Array API supports `nxp.astype(x, dtype)`, using this method causes precision
+    # errors with Jax.
```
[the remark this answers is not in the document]
alxmrs: I tried to use `nxp.astype()` just now and found some friction. Searching SO, it seems like the way I'm creating numpy scalars from python scalars is actually recommended, even without concern for JAX: https://stackoverflow.com/a/66446491/29182710

I'll delete the comment for now; I think this is still a good way to go about it.

alxmrs:
Hey @tomwhite -- I think I may be getting ahead of myself here. I was just adding default_dtypes to `nan_functions.py` now, but it occurs to me: these defaults seem to promote the precision, not reduce it. Is this intentional? If so, I think that's the last place where default dtypes may be needed. 

For now, I'll add a commit with default_dtypes with the intention of removing it should it no longer be necessary.

cubed/array_api/creation_functions.py
```diff
@@ -319,3 +341,12 @@ def _like_args(x, dtype=None, device=None, chunks=None, spec=None):
     if spec is None:
         spec = x.spec
     return dict(shape=x.shape, dtype=dtype, device=device, chunks=chunks, spec=spec)
+
+
+def _closest_default_dtype(dtype, *, device=None):
+    """Returns a dtype most similar to a default (likely changing percision)."""
```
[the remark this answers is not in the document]
alxmrs: It looks like this wasn't needed after all! Thanks for calling it out. Probably it came up simply due to the order I added features/tried to get tests to pass.

cubed/array_api/dtypes.py
```diff
@@ -86,3 +87,35 @@
     "complex floating-point": _complex_floating_dtypes,
     "floating-point": _floating_dtypes,
 }
+
+
+# A Cubed-specific utility.
+def _validate_and_define_dtype(x, dtype=None, *, allowed_dtypes=("numeric",), fname=None, device=None):
+    """Ensure the input dtype is allowed. If it's None, provide a good default dtype."""
```
tomwhite, line 94: Can we rename this something more concrete, like `_upcast_integral_dtypes`? It's used by `sum` and `prod` to ensure that the integral type is wide enough to avoid overflow (see dtype param doc for [`xp.sum`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.sum.html#array_api.sum), and similarly for `xp.prod`).

cubed/array_api/dtypes.py
```diff
@@ -86,3 +87,35 @@
...
+
+    # Choose a good default dtype, when None
+    if dtype is None:
+        if x.dtype in _boolean_dtypes:
+            dtype = dtypes["integral"]
+        elif x.dtype in _signed_integer_dtypes:
+            dtype = dtypes["integral"]
+        elif x.dtype in _unsigned_integer_dtypes:
+            # Type arithemetic to produce an unsinged integer dtype at the same default precision.
+            dtype = nxp.dtype(dtypes["integral"].str.replace("i", "u"))
```
tomwhite, line 113: According to the spec, dtypes are not required to have a `str` attribute, or a str constructor - so this may not work for all implementations. However, I don't know another way of doing it!

cubed/array_api/dtypes.py
```diff
@@ -86,3 +87,35 @@
...
+        elif x.dtype == _complex_floating_dtypes:
+            dtype = dtypes["complex floating"]
+        elif x.dtype == _real_floating_dtypes:
+            dtype = dtypes["real floating"]
```
tomwhite, line 117: We should remove the two cases for complex and real floating, since `sum` and `prod` should ***not*** upcast in this case (see https://data-apis.org/array-api/latest/API_specification/generated/array_api.sum.html#array_api.sum).

Also, `nansum` should be the same as `sum`. It's not in the spec, but NumPy behaves the same (version 2.1.1):

```python
>>> np.nansum(np.array([1.0, 2.0], dtype=np.float32))
np.float32(3.0)
```

cubed/array_api/dtypes.py
```diff
@@ -86,3 +87,35 @@
...
+        raise TypeError(errmsg)
+
+    # Choose a good default dtype, when None
+    if dtype is None:
+        if x.dtype in _boolean_dtypes:
+            dtype = dtypes["integral"]
+        elif x.dtype in _signed_integer_dtypes:
+            dtype = dtypes["integral"]
+        elif x.dtype in _unsigned_integer_dtypes:
+            # Type arithemetic to produce an unsinged integer dtype at the same default precision.
```
tomwhite, line 112:
```suggestion
            # Type arithmetic to produce an unsigned integer dtype at the same default precision.
```

tomwhite (commented):
Getting closer! I added a few more comments.

cubed/array_api/dtypes.py
alxmrs: Really good point! I found a way to do it using this function: 
https://numpy.org/doc/2.1/reference/generated/numpy.iinfo.html

PR #671, opened 2025-01-15
Adding pre-commit (for linting) to test dependencies.

### LINKED ISSUE #668: Add linting to CI
We have a pre-commit config file, but it's not used in CI.

alxmrs:
Fixes #668.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,4 +1,4 @@
-name: Linting
+name: pre-commit
 
 on:
   push:
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,20 @@
+name: Linting
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  checks:
+    name: pre-commit hooks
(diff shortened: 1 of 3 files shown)


tomwhite:
I'm not sure we need to add pre-commit for testing do we? Perhaps we could use https://pre-commit.ci/, or maybe we could add another GH action (or extend/modify the mypy one). Might be worth seeing what other projects do.

alxmrs:
Hmm. Maybe a better fit would be to add an extra just for linting? Then, when we eventually add this to CI, the lint script can run in it's own check? WDYT?

tomwhite:
If possible, I'd copy what another project is doing rather than coming up with a new way of doing things.

A quick look shows:
* Bio2zarr has a CI job to [run pre-commit](https://github.com/sgkit-dev/bio2zarr/blob/main/.github/workflows/ci.yml#L10-L18)
* Xarray uses precommit.ci (see [e.g. here](https://github.com/pydata/xarray/pull/9883/commits/bedf6eba467ab0c71e0ad42b395c72b5a2a9101b)), as does [VirtualiZarr](https://github.com/zarr-developers/VirtualiZarr/pull/322)

tomwhite:
Thanks @alxmrs!

Issue #467, opened 2024-05-21
HPC executor


tomwhite:
Some possibilities:
* Job arrays: https://slurm.schedmd.com/job_array.html (pointed out by @jeromekelleher)
* funcX: https://funcx.org/
* Parsl: https://parsl-project.org/

cc @TomNicholas

TomNicholas:
The main NCAR machine I have access to uses PBS - perhaps there is an equivalent of slurm job arrays for PBS?

TomNicholas:
I've been looking into this more, and all of these look potentially interesting to us!

> perhaps there is an equivalent of slurm job arrays for PBS?

[There is, and it looks very similar](https://centers.hpc.mil/users/advancedTopics/Job_Arrays.html). Perhaps a `PBSExecutor` and `SlurmExecutor` could both inherit from an abstract `JobArrayExecutor`...

I actually started sketching this out, borrowing heavily from `dask-jobqueue`, which defines [common base classes](https://github.com/dask/dask-jobqueue/blob/a35ac91bbf616ef193327c72fe0d4fced80e768b/dask_jobqueue/pbs.py#L38) for different HPC cluster classes. 

But got stuck when I realised that:

1. I need a way to propagate both the environment and the context of an arbitrary function (e.g. in `apply_gufunc`) to the bash scripts that are each job,
2. It might be really slow to have to wait for the queuing system to start running the tasks between every stage. Ideally once we have been allocated resources we don't really want to release them - with cloud serverless this doesn't matter so much as the time between requesting a worker and getting one is extremely short, but on a HPC queue it could literally be hours. A lot of real-world workloads are going to require the most resources at the start - could we imagine that e.g. a reduction with 3 rounds gets 100 workers, then keeps hold of 10 for the next round, then keeps hold of 1 for the final round?

> funcX: https://funcx.org/

funcX has recently become [Globus Compute](https://www.globus.org/compute), a new offering from the same people who run the widely-used Globus file transfer service. The Globus Compute [`Executor` class](https://globus-compute.readthedocs.io/en/latest/executor.html) is a subclass of Python’s [concurrent.futures.Executor](https://docs.python.org/3/library/concurrent.futures.html#executor-objects), so presumably could be used similarly to the existing `ProcessesExecutor`?.

[...]

PR #689, opened 2025-02-04
Globus Compute executor

tomwhite:
This is a proof-of-concept for #467. It takes advantage of the fact that Globus Compute has [an implementation of Python's `concurrent.futures.Executor`](https://globus-compute.readthedocs.io/en/latest/executor.html).

I tested it by running a Globus Compute endpoint locally, by following https://globus-compute.readthedocs.io/en/latest/endpoints/endpoints.html.

To set up the [execution environment](https://globus-compute.readthedocs.io/en/latest/endpoints/endpoints.html#python-based-environment-isolation), I created a `globus-compute-worker` conda env, and ran `pip install cubed globus-compute-endpoint` in it.

Here's the `config.yaml` for the endpoint (stored in `~/.globus_compute/<endpoint-name>/`):

```
display_name: null
engine:
  max_workers_per_node: 1
  provider:
    init_blocks: 1
    max_blocks: 1
    min_blocks: 0
    type: LocalProvider
    worker_init: |
      conda activate globus-compute-worker
      conda env list
  type: GlobusComputeEngine
```

Here's my cubed.yaml (stored in `$(pwd)/globus-compute`):
```
spec:
  allowed_mem: "2GB"
  executor_name: "globus-compute"
  executor_options:
    endpoint_id: '1345cf05-8762-4c24-8def-a131252cf2de'
```

Then I was successfully able to run

```
CUBED_CONFIG=$(pwd)/globus-compute python add-asarray.py
```

### DIFF
--- a/cubed/runtime/create.py
+++ b/cubed/runtime/create.py
@@ -18,6 +18,10 @@ def create_executor(name: str, executor_options: Optional[dict] = None) -> Execu
         from cubed.runtime.executors.dask import DaskExecutor
 
         return DaskExecutor(**executor_options)
+    elif name == "globus-compute":
+        from cubed.runtime.executors.globus_compute import GlobusComputeExecutor
+
+        return GlobusComputeExecutor(**executor_options)
     elif name == "lithops":
         from cubed.runtime.executors.lithops import LithopsExecutor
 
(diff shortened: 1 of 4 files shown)


tomwhite:
I'd like to write a CI workflow to run this but I'm not sure if it's possible to run a local endpoint without having to register it. (We could do that and store the credentials in GH Actions secrets, but that's more involved.)

PR #697, opened 2025-02-17
Mem utilization tests fail on zarr-python v3

tomwhite:
This updates the slow tests (that run test_mem_utilization.py) to run on both zarr-python v2 and v3.

### DIFF
--- a/.github/workflows/slow-tests.yml
+++ b/.github/workflows/slow-tests.yml
@@ -18,7 +18,8 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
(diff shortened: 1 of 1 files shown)


tomwhite:
The reason for the failures is because Zarr Python v3 makes more memory copies than v2 in some cases, see https://github.com/tomwhite/memray-array

tomwhite:
I'm going to merge this now. Note that this runs the mem utilization tests only on Zarr v2 for the time being. Most work on Zarr v3, but there is still work to get them all passing.

The slow tests are failing due to a problem in `cumulative_sum` that I will fix next.

PR #700, opened 2025-03-04
Multistage rechunking

### LINKED ISSUE #650: Multi-stage rechunk
The rechunker algorithm has support for [multi-stage rechunking](https://github.com/pangeo-data/rechunker/pull/89), which reduces the number of IO operations.

It would be good to enable this in Cubed `rechunk`. The simplest way to do this would be to expose the `min_mem` parameter, but it might be interesting to see if there is a way of automatically finding the number of stages that optimizes for something we care about (e.g. min IO ops).

tomwhite:
Fixes #650


tomwhite:
The new implementation is off by default, so I'm going to merge this while I continue testing.

PR #704, opened 2025-03-15
Spark executor

songhan89:
This PR is for https://github.com/cubed-dev/cubed/issues/499


.github/workflows/mypy.yml
```diff
@@ -42,7 +42,7 @@ jobs:
 
       - name: Install mypy
         run: |
-          python -m pip install mypy
+          python -m pip install mypy pyspark
```
tomwhite, line 45: Not sure why this is needed. Is adding an entry for `ignore_missing_imports` in setup.cfg not sufficient?

.gitignore
```diff
@@ -137,3 +137,6 @@ dmypy.json
 # Cubed
 .lithops_config
 examples/cubed.svg
+
+# Pytest artefacts
```
tomwhite, line 141: The `history` directory is actually created by Cubed's `HistoryCallback` so you can remove this line.

cubed/runtime/executors/spark.py
```diff
@@ -0,0 +1,93 @@
...
+        resume: Optional[bool] = None,
+        spec: Optional[Spec] = None,
+        compute_id: Optional[str] = None,
+        **kwargs: Any,
+    ):
+        # Store callbacks for later use during computation
+        self._callbacks = callbacks
+
+        # Clear any previous operations
+        self._lazy_operations = []
```
tomwhite, line 70: Doesn't seem to be used?

cubed/runtime/executors/spark.py
```diff
@@ -0,0 +1,93 @@
...
+        Convert memory value to a valid Spark memory setting string.
+        Ensures the value is at least MIN_MEMORY_MB.
+
+        Args:
+            memory_value: Memory value (can be int bytes or string like '1g')
+
+        Returns:
+            String memory setting suitable for Spark config (e.g., '512m')
+        """
+        # If it's already a string with a unit, return as is
```
tomwhite, line 39: The value from `spec.allowed_mem` is [always an int](https://github.com/cubed-dev/cubed/blob/2b39036de3c3398480f9cfa7ff2f4523470e7fdd/cubed/spec.py#L61) representing the number of bytes, so you don't need to do this check. You could also annotate `memory_value` in this function as an `int`.

cubed/runtime/executors/spark.py
```diff
@@ -0,0 +1,93 @@
...
+        if isinstance(memory_value, str) and any(
+            memory_value.lower().endswith(unit) for unit in ["k", "m", "g", "t"]
+        ):
+            return memory_value
+
+        # Try to convert to bytes if it's a number
+        try:
+            bytes_value = int(memory_value)
+            # Convert to MB and ensure minimum threshold
+            mb_value = max(self.MIN_MEMORY_MB, bytes_value // (1024 * 1024))
```
tomwhite, line 49: Strictly speaking this value is in [MiB](https://en.wikipedia.org/wiki/Byte#Multiple-byte_units), so it would be good to change the comments and the variable naming appropriately.

tomwhite (changes requested):
Thank you for the contribution @songhan89! This looks great - I've made a few minor comments for a few things.

.gitignore
songhan89: Done!

.github/workflows/mypy.yml
songhan89: got it.

cubed/runtime/executors/spark.py
songhan89: Overlooked this. Have removed.

cubed/runtime/executors/spark.py
songhan89: done.

tomwhite (approved):

tomwhite:
Thanks @songhan89 - this is a great addition!

songhan89:
Thanks for the comments and guidance @tomwhite ! Looking forward to contribute more !

tomwhite:
Awesome! It would be great to hear your experience using this on Spark clusters.

PR #713, opened 2025-04-21
Add Databricks setup instructions.

songhan89:


tomwhite (commented):
Thanks for the contribution @songhan89. Can you add this to a new 'databricks' directory under examples (https://github.com/cubed-dev/cubed/tree/main/examples) so it's consistent with the other ones please.

PR #734, opened 2025-06-20
Add MemoryVisualizationCallback 

tomwhite:
This is based on @TomNicholas' work in https://gist.github.com/TomNicholas/c6a28f7c22c6981f75bce280d3e28283, the main difference being that the x-axis is task (number) rather than time, which gives a less distorted plot. (Using time tends to overplot many tasks on top of each other as many tend to start or finish at about the same time.)

Here's an example for the ERA5 rechunking I did in #650:
 
era5-mem-usage

Interestingly, the last stage of rechunking seems to have exceeded the projected memory - but not the allowed memory, so it didn't fail. It would be worth looking into this to see if the projected memory calculation can be improved though.

### DIFF
--- a/cubed/diagnostics/history.py
+++ b/cubed/diagnostics/history.py
@@ -17,6 +17,7 @@ class HistoryCallback(Callback):
                     name=name,
                     op_name=node["op_name"],
                     projected_mem=primitive_op.projected_mem,
+                    allowed_mem=primitive_op.allowed_mem,
                     reserved_mem=primitive_op.reserved_mem,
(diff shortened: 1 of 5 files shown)


TomNicholas:
This looks great!

nit: Is "max peak measured" a tautology?

tomwhite:
> This looks great!

Thanks!

> nit: Is "max peak measured" a tautology?

I don't think so. Each task has a peak memory usage, and we find the maximum across all tasks.

But maybe it's fine to just say "max measured" or "max usage".

PR #737, opened 2025-07-02
Add a `blocks` property to Cubed arrays to allow block-level indexing

tomwhite:
This is similar to Dask and Zarr.


dcherian:
Nice. FWIW I need "outer indexing" of blocks in flox so I added it:
https://github.com/xarray-contrib/flox/blob/b20545b3bd6856529107496b3793134ada57f2c1/flox/core.py#L1713

The core logic is in `_normalize_indexes`, and is more than what you'd need here I think.

Issue #731, opened 2025-06-17
Use zarrs-python for intermediate storage


tomwhite:
We should try the Rust-based [zarrs-python](https://zarrs-python.readthedocs.io/en/latest/) for storing intermediate data to see if it speeds up Cubed. It is installed as a Zarr Python 3 codec pipeline, so should be easy to test.

Thanks to @alxmrs for the suggestion!

Issue #715, opened 2025-06-11
Use obstore for intermediate storage


tomwhite:
Now that Zarr has a [obstore-based store implementation](https://github.com/zarr-developers/zarr-python/releases/tag/v3.0.7), we should try it for storing intermediate data.

PR #751, opened 2025-07-16
build: use hatchling

lgray:
~This is just a suggestion, happy to discuss, but thinking to the long time based on some things we learned in dask in high energy physics. Specifically that our graphs can get very large and a compiled taskgraph api may be something we desire, likewise being able to easily include extensions may help the project in the future. I know this is a premature optimization, but it also doesn't negatively affect anything.~

~For now, running cmake in scikit-build-core is turned off since this repository has no compiled code.
If the project goes with rust, maturin is an easy change from here.~

I have removed the setup.py and setup.cfg, so all configuration is in pyproject.toml and changed the build system to hatchling.
I have also moved cubed's implementation into a `src/` directory so there's no possibility for clobbering packages.

Tests may fail for a bit here where I find all the rough edges from the change.

Curious to hear your thoughts!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,9 @@ MANIFEST
 *.manifest
 *.spec
 
+# scikit-build-core dynamic version
+src/cubed/version.py
+
 # Installer logs
 pip-log.txt
 pip-delete-this-directory.txt
--- a/src/cubed/vendor/dask/array/core.py
+++ b/src/cubed/vendor/dask/array/core.py
@@ -4,7 +4,7 @@ import math
 from numbers import Number
 
 import numpy as np
-from tlz import first, frequencies
+from tlz import frequencies
 
 from ..utils import is_integer, parse_bytes
 
@@ -312,7 +312,7 @@ def auto_chunks(chunks, shape, limit, dtype, previous_chunks=None):
         return tuple(chunks)
 
     if limit is None:
-        limit = "128MiB" # config.get("array.chunk-size")
+        limit = "128MiB"  # config.get("array.chunk-size")
     if isinstance(limit, str):
         limit = parse_bytes(limit)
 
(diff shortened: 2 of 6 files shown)


tomwhite:
Thanks for the PR @lgray - a few comments below.

> Specifically that our graphs can get very large and a compiled taskgraph api may be something we desire

Cubed should be able to help with this since each node in the graph is an array not a chunk.

> For now, running cmake in scikit-build-core is turned off since this repository has no compiled code.

I would rather make this change when it is needed - i.e. alongside the PR that introduces the compiled code.

> If the project goes with rust, maturin is an easy change from here.

I can see some parts of the Cubed stack migrating to Rust in the future. For storage we already have obstore support (#715), and zarrs-python (#731) is on the TODO list. Both of these are simply using Rust-based Python libraries though so don't need any build changes, but we could have Rust code in Cubed for the runtime, primitive (blockwise, rechunk) and core (planning) layers. I think it's natural to make the necessary build changes at the point we introduce the compiled code.

> I have removed the setup.py and setup.cfg, so all configuration is in pyproject.toml.

This is a good change - could you break this off as a separate PR and we'll get it merged.

tomwhite:
> OK - I'll move everything to hatchling in the meantime. Do you have any opinion on `src/` vs the package in the root of the repository?

I don't have a strong opinion on that. @TomNicholas do you have any thoughts?

TomNicholas:
No strong opinion on that either

lgray:
OK - kept the `pyproject.toml`-only changes, used a python-only / more minimal build system. All the code is now beneath `src/`

tomwhite:
I just enabled the CI - many are passing! But looks like [.coveragerc](https://github.com/cubed-dev/cubed/blob/main/.coveragerc) is not being picked up?

lgray:
Thanks! I'll see what I can do to get it all going. It seems to not be picking up the mypy config correctly either.

tomwhite:
No problem @lgray! Moving the `src` dir may require a bit of coordination with other PRs so let me know when you're ready.

I've also been fixing some of the flaky test failures.

PR #772, opened 2025-07-25
Basic support for device arrays

### LINKED ISSUE #750: Experimenting with GPU / cupy
I'm reading up on some related issues (https://github.com/cubed-dev/cubed/issues/304, https://github.com/cubed-dev/cubed/issues/307) but wanted to check on the status of cubed with cupy arrays. https://github.com/TomAugspurger/cubed-cupy has my notes. I needed a couple small tweaks to `cubed` to get things working:

1. Use `array_api_compat.cupy` instead of `array_api_compat.numpy`. The main `cupy` doesn't seem to be compatible with the array api yet (`cupy.bool` vs. `cupy.bool_` is the first issue I hit; haven't looked further yet). edit: https://github.com/cupy/cupy/issues/9249 for that.
2. We don't really ever want to convert from a `cupy.ndarray` to a `numpy.ndarray`, since it's (usually) *not* a simple view on a buffer. `np.asarray(cupy_array)` triggers a copy from the GPU to host memory. I haven't looked into where all this occurs, but as a hack for now I've just opted to not convert to a `numpy.ndarray`.

Those changes are at https://github.com/cubed-dev/cubed/compare/main...TomAugspurger:cubed:twa/cupy-compat?expand=1.

Finally, zarr-python natively supports cupy buffers, so we enable that with `zarr.config.enable_gpu()`. With those changes, I was able to run the examples at https://cubed-dev.github.io/cubed/examples/basic-array-ops.html.

Has anyone else tried tried this combination?

If anyone else is interested in this let me know and I can push things along.

TomAugspurger:
This PR fixes a few issues preventing us from using cupy arrays for basic operations. With the changes on this branch, I can run

```
❯ ZARR_BUFFER="zarr.buffer.gpu.Buffer" ZARR_NDBUFFER="zarr.buffer.gpu.NDBuffer" CUBED_BACKEND_ARRAY_API_MODULE=array_api_compat.cupy pytest cubed/tests/test_array_api.py -vs
```

without error, assuming you have cupy and zarr-python 3.x installed.

## Summary of Changes

Most things worked already. The majority of the changes consist of

- small changes to use the array API rather than `numpy` to allocate data (e.g. `np.empty`)
- changes to `assert_array_equal` to handle host to device transfers for `cupy.ndarray` (you need to explicitly call `.get()` to get a numpy ndarray)
- change `backend_array_to_numpy_array` to preserve cupy arrays: My understanding is that this is primarily used to ensure that zarr gets a NumPy ndarray. zarr-python 3.x has support for device buffers (currently just cupy, but hopefully more in the future). So we want to delay transfering from device to host and let zarr do that.
- skipping certain tests
  - all the process based tests (need to figure out how best to handle this)
  - the `getitem` test using `ndindex`, which AFAICT only supports numpy currently

There's still plenty more to do to make this fully functional, but I think this removes the initial hurdles to using cubed with cupy.

Closes #750

### DIFF
--- a/cubed/backend_array_api.py
+++ b/cubed/backend_array_api.py
@@ -32,6 +32,7 @@ else:
     import array_api_compat.numpy
 
     namespace = array_api_compat.numpy
+    xp_name = "numpy"
 
 
 # These functions to convert to/from backend arrays
--- a/cubed/backend_array_api.py
+++ b/cubed/backend_array_api.py
@@ -41,6 +41,9 @@ else:
 
 
 def backend_array_to_numpy_array(arr):
+    # temporarily disable this. for cupy `arr` and GPU buffers in Zarr, we don't
+    # want to convert to NumPy.
+    return arr
     return np.asarray(arr)
 
 
--- a/cubed/backend_array_api.py
+++ b/cubed/backend_array_api.py
@@ -38,13 +38,21 @@ else:
 # assume that no extra memory is allocated, by using the
 # Python buffer protocol.
 # See https://data-apis.org/array-api/latest/API_specification/generated/array_api.asarray.html
+if "cupy" in namespace.__name__:
+    # zarr-python 3.x natively supports some device buffers (currently just cupy,
+    # but https://github.com/zarr-developers/zarr-python/issues/2658 is expanding the
+    # set). For these backends, we *don't* want to copy to the host.
 
+    def backend_array_to_numpy_array(arr):
+        return arr
 
-def backend_array_to_numpy_array(arr):
-    # temporarily disable this. for cupy `arr` and GPU buffers in Zarr, we don't
-    # want to convert to NumPy.
-    return arr
-    return np.asarray(arr)
+else:
+
+    def backend_array_to_numpy_array(arr):
+        # temporarily disable this. for cupy `arr` and GPU buffers in Zarr, we don't
+        # want to convert to NumPy.
+        # return arr
+        return np.asarray(arr)
 
 
 def numpy_array_to_backend_array(arr, *, dtype=None):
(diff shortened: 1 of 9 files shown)


cubed/backend_array_api.py
```diff
@@ -38,10 +38,21 @@
...
+        return arr
 
-def backend_array_to_numpy_array(arr):
-    return np.asarray(arr)
+else:
+
+    def backend_array_to_numpy_array(arr):
+        # temporarily disable this. for cupy `arr` and GPU buffers in Zarr, we don't
+        # want to convert to NumPy.
+        # return arr
```
tomwhite, line 54: Remove these three comment lines?

docs/user-guide/gpus.md
```diff
@@ -0,0 +1,24 @@
+# GPU Support
+
+Cubed has experimental support for using GPU-backed ndarrays. With zarr-python's
+[native GPU support], you can load data into GPU memory, perform some cubed
+computation on the GPU, and write the result, while minimzing the number of host
```
tomwhite, line 5:
```suggestion
computation on the GPU, and write the result, while minimizing the number of host
```

tomwhite (commented):
This looks great @TomAugspurger!

Quick comments:
* The mypy failure needs fixing by adding the relevant line here: https://github.com/cubed-dev/cubed/blob/main/setup.cfg#L22
* The JAX tests are now failing because of the change in `_assemble_index_chunk`, which now exposes the fact that JAX can't do in-place mutation in the usual way. Not sure how to fix that yet...

TomAugspurger:
The only failing test is `test_default_spec[lithops]`: https://github.com/cubed-dev/cubed/actions/runs/16546166408/job/46800471136?pr=772#step:6:2279, which fails with

```
        if errors:
>           raise Error(errors)
E           shutil.Error: [('/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/lithops/__pycache__', '/tmp/lithops-runner/lithops/__pycache__', "[Errno 17] File exists: '/tmp/lithops-runner/lithops/__pycache__'")]
```

I'm unfortunately not able to reproduce that locally.

Issue #837, opened 2025-11-21
Ability to visualize plans that exceed memory


tomwhite:
> I think it would be even more useful if you could call `visualize` to get a plan that you can look at even if memory is exceeded. You'd get a warning in that case. And calling `compute` would raise the exception. 

 _Originally posted by @tomwhite in [#568](https://github.com/cubed-dev/cubed/issues/568#issuecomment-3559155003)_

The visualization could mark the offending operations in red.

tomwhite:
Fixed in #846

PR #846, opened 2025-12-01
Visualize plans that exceed memory

neilSchroeder:
I saw #837 and thought it would be a good place for a more meaningful contribution. 

Here's a PR that implements this feature.

memory_exceeded_plan

I've changed how the system handles and reports operations that exceed allowed memory in computation plans. Instead of failing immediately, plans can now be created even if some operations exceed memory limits and errors are raised when execution is attempted. Plans with operations exceeding memory limits are clearly marked with an internal property. The visualization highlights bad operations and issues a warning. Tests have been updated to reflect this new behavior.

### DIFF
--- a/cubed/core/plan.py
+++ b/cubed/core/plan.py
@@ -639,7 +639,9 @@ class FinalizedPlan:
 
         if self._ops_exceeding_memory:
             # Build warning text in red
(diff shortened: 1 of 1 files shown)


tomwhite (approved):
Brilliant! Thanks for working on this @neilSchroeder!

PR #861, opened 2026-01-13
Feature/virtualizarr support

neilSchroeder:
# Virtual Array Support

## What
This is a shot in the dark at attempting to implement `ManifestArray` support in `cubed`.

## Why
There's an ideal workflow that looks like:
1) A file is stored as a NetCDF
2) Open that file as a virtual dataset with VirtualiZarr
3) Manipulate that array with Cubed
4) Write to Zarr

This makes use of the zero-copy functionality of VirtualiZarr and allows for embarassingly parallel workflows that start with NetCDFs and do not need intermediate Zarrs.

## Some Comments
@tomwhite & @TomNicholas I would love your feedback on the implementation here, I sorta hacked this together and tried to get it to work with using a real-world example. It works for my use case now, but I'm not convinced I've come up with a particularly thorough/generalizable implementation.


neilSchroeder:
Hi @TomNicholas, happy to clarify. Here's some pseudo code that I think illustrates the problem:

```python
# Step 1: Create virtual dataset
manifest_dataset = virtualizarr.open_virtual_mfdataset(
    s3_files,  # 8,760 file paths
    parallel="lithops"  # Uses serverless workers to read metadata
)
# Result: manifest_dataset contains only references
#   Example: {"path": "s3://bucket/file.nc", "offset": 1024, "length": 8192}
#   No actual data downloaded

# Step 2: Convert VirtualiZarr manifests to Cubed arrays
temperature = cubed.from_manifest(
    load_chunk=lambda key: fetch_from_s3(manifest[key]),  # Serializable loader
    shape=(8760, 39, 150, 150),
    chunks=(1, 39, 50, 50),
    dtype='float32'
)
# Result: Cubed array with lazy chunk loaders
#   Still no data movement - just a computation plan

# Step 3: Build computation DAG (still lazy)
pressure = cubed.from_manifest(...)
interpolated = cubed_operations.interp_to_pressure_levels(
    temperature, 
    pressure, 
    target_levels=[1000, 850, 500]  # Standard atmospheric levels
)
# Result: Computation graph, no execution yet

# Step 4: Execute on serverless workers (Lithops/Lambda)
interpolated.compute(executor="lithops")
# Result: Each worker:
#   1. Gets chunk key: (time=100, level=slice, y=0:50, x=0:50)
#   2. Calls load_chunk(key) → fetches ONLY that chunk from S3
#   3. Processes chunk → interpolates to pressure levels
#   4. Writes result to Zarr
# Data never passes through client
```

I was getting stuck attempting to go directly from virtual datasets to cubed because of the ManifestArrays, so this is my attempt to solve that issue. I tried to find documentation on how to go about this, but couldn't come up with anything that would let me use lithops to get embarassingly parallel output to zarr.

TomNicholas:
Does that make sense @neilSchroeder ? Did you try the approach I suggested instead?

PR #893, opened 2026-04-08
Add compatibility with Zarr 3.2.0

maxrjones:
We changed some internals when adding variable chunks support in zarr-python. This PR makes cubed compatible with the future zarr-python 3.2.0 release.

### DIFF
--- a/cubed/core/ops.py
+++ b/cubed/core/ops.py
@@ -625,10 +625,20 @@ def elemwise(func, *args: "Array", dtype=None) -> "Array":
 
 def _create_zarr_indexer(selection, shape, chunks):
     if zarr.__version__[0] == "3":
-        from zarr.core.chunk_grids import RegularChunkGrid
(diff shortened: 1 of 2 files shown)


tomwhite:
Thanks @maxrjones! It looks like we can merge this now without waiting for Zarr 3.2.0, right?

maxrjones:
> Thanks @maxrjones! It looks like we can merge this now without waiting for Zarr 3.2.0, right?

hey, sorry for me slow reply. Yes, this should be safe to merge in advance of 3.2.0. We took a similar approach in Virtualizarr. 

We're hoping to make some of this API public in the future, which should allow you to rely on it with more stability provided.

rsignell:
Yay!   Might this warrant a new release?

tomwhite:
> Yay! Might this warrant a new release?

Yes, I'll try and do one soon.

PR #894, opened 2026-04-10
Implement `__setitem__` for Cubed arrays

tomwhite:


tomwhite:
Closing this since it's the wrong approach. Cubed array's are immutable, so it would be better to implement a JAX-like update: `x = x.at[idx].set(y)`, see #906.

Issue #420, opened 2024-03-11
Failing tensordot test


tomwhite:
From https://github.com/cubed-dev/cubed/actions/runs/8231307282/job/22506341591?pr=419:

```
FAILED array_api_tests/test_linalg.py::test_tensordot - AssertionError: out.shape=(1, 0), but should be (0, 0) [tensordot( (0, 0) . (0, 0) )]
Falsifying example: test_tensordot(
    x1=cubed.Array<array-4451, shape=(0, 0), dtype=uint8, chunks=((0,), (0,))>,
    x2=cubed.Array<array-4452, shape=(0, 0), dtype=uint8, chunks=((0,), (0,))>,
    kw={'axes': ((-2,), (0,))},
)

You can reproduce this example by temporarily adding @reproduce_failure('6.99.2', b'AXicY2BgYARCCMEApZkAAGAACQ==') as a decorator on your test case
```

There was a similar failure in running #402.

tomwhite:
Fixed in #907

PR #907, opened 2026-05-26
Array API test fixes

tomwhite:
A bunch of corner cases found by running Hypothesis with default max examples (100)


tomwhite:
Fixes #420

PR #927, opened 2026-06-08
Rechunk store inefficiency fix

### LINKED ISSUE #877: Storing rechunked array uses too many tasks
For example, I noticed in #859 the final `_store_array` op has >200K tasks, while the one before it (`rechunk`) has only 1560 tasks, as shown in the plan below. These two ops should be fused and should only have 1560 tasks.

Image

What has happened is that the `_store_array` op is based on the number of chunks in the final array, rather than the number of write chunks in the final rechunk stage.

tomwhite:
Fixes #877

### DIFF
--- a/cubed/core/ops.py
+++ b/cubed/core/ops.py
@@ -282,6 +282,8 @@ def _store_array(
                     writes_map = pipeline.config.writes_map
                     if source.name in writes_map:
                         writes_map[source.name].array = target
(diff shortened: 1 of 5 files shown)


tomwhite:
This fix makes a few trade offs (a couple of dataclasses need to be mutated; rewiring the DAG is delicate) - but it's a fairly straightforward fix for the very problematic behaviour described in #877 - so it's worth merging I think.

